PR #13, opened 2019-11-22
remove scroll bars from side navigation when not needed

Sparclex:
Currently the scroll bars for the navigation are always shown even if there is nothing to scroll (might only be visible on windows)

image
image

With this change the scroll bars only appear when there is content to scroll.


jackmcdade:
Yeah that's a Windows thing, but if it works, it works!

PR #22, opened 2019-12-07
Use slug as arg in Utility::make function

mattrothenberg:
The string `french fries` causes `UtilitiesController` to blow up, as its expecting a valid URL entity when it calls the `request->segment` method.


jackmcdade:
🔥 🚒 Thank you!

PR #25, opened 2019-12-09
Remove semicolon from inside array

damcclean:
My code didn't work when I had the semi-colon here because it's inside an array.


jackmcdade:
Nice catch!

PR #26, opened 2019-12-12
Fixed text-input usage

damcclean:
I've fixed the usage of the `text-input` component by adding an end tag to the documentation because you can't have a starting tag without an ending tag (most of the time anyway)


jasonvarga:
It should just be a self closing tag, it's missing a `/>`

damcclean:
yep... let me update that

damcclean:
there we go! fixed. 

I actually realised I had fixed it wrong and was about to change it when I saw you comment 😂

PR #27, opened 2019-12-13
Fix reference to mail.php location

andrewcgregory:
Fixed reference to `mail.php` location and added internal link for env.


jasonvarga:
Thanks!

PR #35, opened 2019-12-17
Typo fix

calebporzio:
I think?


jasonvarga:
Seems better to me, thanks for your efforts!

jackmcdade:
Ah poo, I meant `effortizationification` but this is probs ok too

PR #37, opened 2019-12-19
Add Nav namespace

jonassiewertsen:
I think this would make it clearer which namespace to use, if you arent familar with Laravel.

Maybe it should be added even more often?


jackmcdade:
Was resolved outside the PR. Thanks for the heads up!

PR #39, opened 2019-12-21
Fix broken link to TipTap site

crgeary:
The link to TipTap [on the bard docs](https://statamic.dev/fieldtypes/bard#extending-bard) had an incorrect URL.

I've changed it from `.com` to `.io`.


jackmcdade:
Thanks!

PR #40, opened 2019-12-21
Add a description on how to add views to your Addon

jonassiewertsen:
A grammar check would be needed :-)

### DIFF
--- a/content/collections/extending-docs/addons.md
+++ b/content/collections/extending-docs/addons.md
@@ -189,6 +189,49 @@ protected $commands = [
 
 ```
 
+## Registering Views
+
+If you want to use your own views, remember to register them in your Service provider:
+
+```php
+
+class ServiceProvider extends AddonServiceProvider
+{
+    public function boot()
+    {
+        parent::boot();
+
+        $this->loadViewsFrom(__DIR__.'/../resources/views/', 'YourAddonName');
+    }
+}
+```
+
+The path `/resources/views/` may be customized as you as you need, but saving your views in the resources/views folder i common practice. 
+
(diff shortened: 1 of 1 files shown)


jasonvarga:
I think it would be way more useful for us to just wire this up to be automatic.

Behind the scenes we could call `$this->loadViewsFrom('path/to/their/resources/views/', 'addon-name');`

Where `addon-name` would be the second part of their composer package (eg. `vendor/addon-name`)

If they have a `resources/views` directory, they should just be able to assume `view('addon-name::myview')` would work.

jonassiewertsen:
Personally, I would enjoy it the "Laravel way" without any automation. 

Especially in custom addons, it's nice to stay in control. 

Let me know if you plan to automate it. I am more than happy to adjust the description!

jasonvarga:
Already automated (https://github.com/statamic/cms/commit/c8c7587bb4ba8b47abc73f889cd7a8a5d07b38a4), and docs done (https://github.com/statamic/docs/commit/14311273a3c3a184cd39da258a7019eea43cd74a). Thanks for offering though!

You're free to continue to do what you're doing though, if you want to put your views somewhere non-standard or use a different namespace.

jonassiewertsen:
Ohh, I did not know about the automation. 

It wasn't there in Dec 2019 :-)

Thanks!

jasonvarga:
It wasn't. I just added it today!

jasonvarga:
And now you can customize the namespace, if that's one of the reasons you'd want more control. (20698303)

jonassiewertsen:
That looks good Jason!

PR #44, opened 2019-12-29
Change "to Markdown" to "to HTML"

owenconti:
I _think_ this should be changed to "HTML" instead of "Markdown", since the content is already written in Markdown.


jasonvarga:
Indeed. Thanks!

PR #45, opened 2020-01-01
Fix the path for global set

MohammadAbusaleh:
Fixed the path for the global set for multiple sites from structure to global


jackmcdade:
Thanks!

PR #46, opened 2020-01-01
Update use Modifier class

jcohlmeyer:
Update `use Statamic\Extend\Modifier;` to `use Statamic\Modifiers\Modifier;`


jackmcdade:
Thanks!

PR #47, opened 2020-01-03
Fixed link to addon docs

damcclean:
I've fixed the link to the addon docs from the Tags page of the documentation. It was previously 404ing.


jackmcdade:
Thanks!

PR #48, opened 2020-01-03
It's make:tag, not make:tags

damcclean:
I tried to run make:tags but the command doesn't actually exist, where as make:tag does.

(sorry for the 2nd pr to the same doc within 2 minutes 😂)


jackmcdade:
Good catch.

jasonvarga:
💡 We could change the command to `make:tags` (plural) and then `make:tag` would also work because it autocompletes.

jesseleite:
Meh, I kinda like how everything is singular, but just my 2 cents!

Screen Shot on 2020-01-06 at 15-20-05

PR #50, opened 2020-01-03
Fix another documentation link 

damcclean:
I've fixed another documentation link that was previously 404ing.


jackmcdade:
Thanks!

PR #51, opened 2020-01-06
fixed vue.js link in frontend-overview docs

robertguss:
The vue.js link in frontend-overview was linking to react's website, so I replaced it with the correct URL.


jackmcdade:
Hah, good one! Thanks :)

PR #57, opened 2020-01-19
Updated Requirements.md to add a windows option

mwesten:
Added the Laragon software as more flexible alternative to WAMP. Also added the description of adding a Quick App to the Laragon menu, so you instantly can install a Statamic3 site.


jackmcdade:
Thanks! I reworded this but the mention of Laragon was a good one.

PR #66, opened 2020-02-14
Update configuration.md - Typos and Suggestions

cnc137:
I fixed a few typos on this page.

I also submitted the following https://github.com/statamic/cms/issues/1358 suggestion that was closed, but it might be helpful to document which environment variables are required for Statamic and which ones are optional on this page. 

- Why are the database variables included in the .env.example file when they are not needed for a flat file system?
- Do I need the Redis variables to make a basic Statamic site work?
- What are the Pusher variables used for? Do I need to set up a https://pusher.com account and get an API key?

The mail variables made sense and I got those up and running and tested with https://mailtrap.io and https://elasticemail.com, but some of the other ones in the .env file will require additional research outside of the Statamic documentation to understand.

If you are going to use the standard .env file from Laravel, then I think you need to explain what is and isn't needed in it for Statamic. I looked at the file and questioned what I needed and what I didn't need.

I have been trying to learn Statamic and Laravel and it would be nice to know what Laravel things I need to worry about and which ones I can ignore for now while getting Statamic up and running.

### DIFF
--- a/content/collections/docs/configuration.md
+++ b/content/collections/docs/configuration.md
@@ -62,7 +62,7 @@ APP_NAME="New Statamic Site"
 
 ### Retrieving Environment Variables
 
-All of the variables listed in this file are available in your config files by using the `env()` helper. The optional second argument lets pass a default value in case you don't have the environment variable set.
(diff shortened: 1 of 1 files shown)


jackmcdade:
Thanks for the suggestions. I'm torn between preserving the standard Laravel experience and fine-tuning the Statamic one. I'm leaning towards the latter.

PR #68, opened 2020-02-16
Update bard.md

sg-modlab:
If I am following v3 will not have another RTE other than Bard so thinking there might be some folks who might use without any sets. While I stumbled upon this for not so thoughtful reasons, thought I'd share forward for consideration if indeed my understanding is accurate.


jackmcdade:
Yes, you’re exactly right. Good call!

PR #72, opened 2020-02-22
Make the docs vegan friendly

tao:
Removed the duplicate Antlers output and made the documentation vegan friendly.

### DIFF
--- a/content/collections/extending-docs/modifiers.md
+++ b/content/collections/extending-docs/modifiers.md
@@ -107,7 +107,7 @@ Given the following data:
 
 ``` yaml
 times: 5
-thing: Bacon
+thing: Pizza
 ```
 
 And template:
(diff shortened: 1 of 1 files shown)


jasonvarga:
Can't we compromise? 

image

tao:
Pizza is a compromise compared to Kale ;) and personally Pizza reminds me a lot more of the Statamic 80’s / early 90’s / futuristic / hacker vibe than Bacon does.

alexharrell:
I agree with pizza, as a bacon lover, I would be offended by any "Facon".

jasonvarga:
You're right. Pizza can be enjoyed by everyone. Meat lovers, vegans, and turtles.

jackmcdade:
But what about the gluten?

jesseleite:
I’m allergic to the dairy, can we compromise with ‘vitamin water’?

jackmcdade:
What about essential oils?

tao:
<3 is shaped like pizza for a reason :smile:

PR #79, opened 2020-02-27
Update search.md

jmalko:
Add 'no_results' condition in order to complete the basic search results sample.  It is easy to get '500' errors when changing the result output when there are no actual results.  This was present in v2 docs, helped me to fix my code.


jackmcdade:
You’re right. I get all happy by nice and concise code examples but the detail is better for everyone.

jmalko:
Thanks Jack.  You're right.

PR #84, opened 2020-03-18
Update require script for debug bar

lbattaglioli2000:
I would've gotten away with it if it weren't for those meddling... hyphens?


jackmcdade:
Thanks Luigi!

PR #93, opened 2020-03-29
Make installation.md clearer for newbies

rsimplicio:
Add details to the installation step to make clear that my-site is going to be the name of the new project/directory, and remind newcomers to Composer to change into their new project directory before attempting to update to prevent 'Composer could not find a composer.json file' errors.


jackmcdade:
Thanks for the BR! I worded it differently but the gist is in there.

PR #101, opened 2020-04-11
Path incorrect for config app. This does not work.

icemancast:
Tried the suggested and it did not, then realized the path noted for the config is incorrect.

### DIFF
--- a/content/collections/docs/configuration.md
+++ b/content/collections/docs/configuration.md
@@ -65,7 +65,7 @@ APP_NAME="New Statamic Site"
 All of the variables listed in this file are available in your config files by using the `env()` helper function. An optional second argument allows you to pass a default value.
 
 ``` php
-// config/statamic/app.php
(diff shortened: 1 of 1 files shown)


jasonvarga:
The path was correct. Perhaps you can explain what wasn’t working for you?

icemancast:
One there is not config/statamic/app.php file in v3 using beta 22 so I created one. After that I tried using the suggest {{ config:app:awesome }} and it doesn't not reference the path described. 

Then tested the addition of the value inside config/app.php with the {{ config:app:awesome }} template tag and that works perfectly.

jasonvarga:
The environment variables section applies to config files in general. Not specific to the ones in the statamic subdirectory.

icemancast:
I understand that so noted above the config tag is: `config/statamic/app.php` which is statamic specific. It caused some confusion if that makes sense.

PR #103, opened 2020-04-17
Update class used to register custom user drivers

imliam:
If I'm not mistaken, these parts of the docs are outdated and incorrect. The `Statamic\Facades\User` class returns an instance of the existing User repository in the container, not the manager needed to register a new driver as you'd anticipate. Additionally. calling the `repository` method on the manager as previously suggested doesn't actually do what you'd expect and add the implementation for the config to pick up later on - you instead need to call the `extend` method.

I'm not 100% certain about this, but from testing and looking around, the change I made in this pull request seems to be correct.

Additionally, this change shows the example code is in the `register` function so people know to put it there instead of `boot` in the service provider.


jasonvarga:
I think this is a code issue, not a docs issue. I'd want it to work the way it's documented.

PR #109, opened 2020-04-28
Add underscore and antlers filetype

damcclean:
When you don't use the underscore and `.antlers.html` filetype then the partial contents don't get picked up.

I've updated the documentation to reflect this.

### DIFF
--- a/content/collections/fieldtypes/bard.md
+++ b/content/collections/fieldtypes/bard.md
@@ -127,11 +127,11 @@ An alternative approach (for those who like DRY or small templates) is to create
 ```language-files
 partials/
 |-- sets/
-|   |-- gallery.html
-|   |-- image.html
-|   |-- poll.html
-|   |-- text.html
-|   |-- video.html
+|   |-- _gallery.antlers.html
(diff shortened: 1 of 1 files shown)


content/collections/fieldtypes/bard.md
```diff
@@ -127,11 +127,11 @@ An alternative approach (for those who like DRY or small templates) is to create
...
-|   |-- gallery.html
-|   |-- image.html
-|   |-- poll.html
-|   |-- text.html
-|   |-- video.html
+|   |-- _gallery.antlers.html
+|   |-- _image.antlers.html
+|   |-- _poll.antlers.html
+|   |-- _text.antlers.html
+|   |-- _video.antlers.html
```
jasonvarga, line 134: We'll want `.antlers`, (although just `.html` should technically work too, it just wouldn't be antlers), but the underscore shouldn't matter.

Can you remove the underscores?

If it's really not working for you without underscores, can you open an issue in the cms repo instead?

damcclean: Done!

PR #114, opened 2020-05-02
publish session tags so that folks know this tag exists

edalzell:
I needed this tag last night (I checked the docs first) to debug some stuff so I wrote it, then discovered this tag does exist but is not `published`. I suspect it was missed when you did your `status` thing.

Hoping you'll accept this so others can see this exists, even if the docs aren't ready yet.


jackmcdade:
Good idea 👌

PR #116, opened 2020-05-06
Added hook description and prerequisites

chrislow:
Help describe hooks more detailed and give the users a more simple guide on how to use them.


jasonvarga:
Hooks are exclusively javascript, and only available in the control panel. I'll tweak that. Thanks for the PR!

PR #119, opened 2020-05-11
removed doubled resources folder

goellner:


jackmcdade:
Nice catch, thank you!

PR #123, opened 2020-05-17
Move words into correct order.

Konafets:


jackmcdade:
lol nice one, thanks!

PR #126, opened 2020-06-03
Episodes collection structure example

mikemartin:


jackmcdade:
I'm not quite sure I understand this one...

mikemartin:
@jackmcdade This might be a work around, but suppose you have a `TV Show` structured collection that is mounted to the url `/tv-show` and has a depth of 2. You might try to use the route `/tv-show/{ parent_uri }/{ slug }`. This results in the following urls for each depth:

depth 1: `/tv-show/season-one`
depth 2: `/tv-show/tv-show/season-one/episode-one`

The condition is used to remove the prepended slug when the `parent_uri` is added.

jackmcdade:
Ahhh, I see. Fair enough! Maybe something a bit more terse? Can you test if this works?

```
route: {{ depth == 1 ??= mount }}{{ parent_uri }}/{{ slug }}
```

jasonvarga:
Being able to slap a prefix in the url for the top level items was definitely intentional. Good to have this as an example, but maybe it could use some more words :)

mikemartin:
@jackmcdade Nice one but doesn't work yet. https://flareapp.io/share/x5MWjDPk

jackmcdade:
I think you might be missing a space before the `??=` bit.

mikemartin:
@jackmcdade Don't think so. Maybe it gets removed?
Screen Shot 2020-06-04 at 9 34 41 am

jackmcdade:
Huh, weird. The error was a bit bizarre. I guess the full parser isn't kicking in here.

mikemartin:
@jackmcdade @jasonvarga  Will wait for a fix for this, but in the meantime I've added a hot tip to explain the route condition. 

https://github.com/statamic/docs/pull/126/files

jasonvarga:
Sorry for the delay here. We ended up going with a slightly different example. 0440332b5d16aefee3ed69ce205503e7d28eb5f3

The PR is appreciated!

PR #131, opened 2020-06-15
Update rest-api documentation

benfurfie:
Grammar standards – API should probably be uppercase as it is on the rest of the page.


jasonvarga:
Maybe it should be `api` (in code). I think that was probably the intention. It's the `api` file.

PR #132, opened 2020-06-18
Add site variables with the new syntax

Konafets:


jackmcdade:
Thank you for this!  I'm wondering if it would be cleaner to just have one page for the `site` variable because it's just an array with keys you can access, not 4 separate variables. What say you @Konafets?

Konafets:
I would say thats sounds like a cleaner approach. I just updated my PR. 
One thing I did not get is, how the sorting of the entries here https://statamic.dev/variables#system-variables is done. The `site` variable is at the very end of the table on my instance of the documentation. Any hints @jackmcdade?

jasonvarga:
This was added already, sorry about that. Appreciate the PR though!

PR #136, opened 2020-06-21
Update config file path

damcclean:
I've seen it a few times, people have thought there is an app.php file within the Statamic file but they can't see it in their local install.


rosswintle:
+1 for this. Tripped me up. Thanks @damcclean

PR #141, opened 2020-06-29
site_url now site:url

rosswintle:
This tripped me up today. I think this changed in [beta32](https://github.com/statamic/cms/releases/tag/v3.0.0-beta.32).

I've submitted this PR for `site:url`, but the docs should probably be reviewed for other changes. This may not even be a correct PR - you may have more substantial work to do around the `site` variables.


jasonvarga:
I've removed the `site_url` page and replaced it with `site` with more details.

https://statamic.dev/variables/site

Thanks for pointing it out!

PR #142, opened 2020-07-04
Adding the word 'if' to Extending Statamic page intro

paulstone:
Adding the word 'if' to:

"even if you had no intention of distributing it"


jackmcdade:
Nice one, thank you!

PR #146, opened 2020-07-12
User profile pages, by username

edalzell:
Also updated the examples to better reflect how the tag works. For example the previous route only worked if an `id` was passed.

Documents https://github.com/statamic/cms/pull/2071


jackmcdade:
Thanks!

PR #149, opened 2020-07-16
Update forms.md

benfurfie:
This form example – much like school and life – will influence the way people build forms for the rest of their Statamic careers (or until a security team smacks their wrists).

As such, it makes sense to be responsible and show them an example of how to set up a form that doesn't re-render form fields after a successful submission (and in doing so will help stop people spamming their forms, which is totally not rad).

Basically, on page load, is this a successful submit? If so, show the success message and don't render anything else.

If not, were there any errors? If so, render them. Otherwise, render the form.

Also, this is entirely to brand standards and, as much as it pained me, uses that /if approach that makes me want to 🤮. Love you guys really.

### DIFF
--- a/content/collections/docs/forms.md
+++ b/content/collections/docs/forms.md
@@ -72,35 +72,37 @@ This example dynamically renders each input's HTML. You could alternatively writ
 
 ```
 {{ form:super_fans }}
-    // show any errors here
-    {{ if errors }}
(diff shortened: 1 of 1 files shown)


jackmcdade:
Thank you for doing that which you hate. You shall be rewarded in due time. ❤️

PR #155, opened 2020-07-25
Vendor Name is not part of public path 

damcclean:
The public path for vendor assets in Statamic no longer contains the vendor's name, just the name of the package, at least this is how I seem to have it setup in my apps.


jasonvarga:
That’s intentional, it’s a Laravel convention.

jasonvarga:
Sorry I’m an idiot, I thought this was the cms repo and you were reporting a bug. Thanks!

PR #161, opened 2020-08-01
Update user:reset_password_form example

StefanSchuechl:
The Statamic\Auth\ResetsPasswords trait validates against an email field.
This should be reflected on the frontend tag example in the docs.


jackmcdade:
Thanks!

PR #162, opened 2020-08-01
Use correct indentation

stancl:
Real men indent their PHP with 4 spaces.

Now it's consistent with the rest of the code samples on the page.


jackmcdade:
😎

PR #164, opened 2020-08-05
Add specifications to inject view models

jonassiewertsen:
I am not sure if this is the best explanation. Still, we should definitely add an example on how to `inject` a view model.


jasonvarga:
Rather than a separate section, I just changed the example to use a collection since that's probably more useful anyway.

PR #170, opened 2020-08-17
Adding hint when using sqlite as db driver

Naoray:


jackmcdade:
Great addition, thank you!

PR #172, opened 2020-08-18
Glide does support the webp format

jonassiewertsen:


jackmcdade:
Thanks! I think that might be a relatively recent addition because I don’t remember that :)

PR #173, opened 2020-08-19
Fix incorrect response data

davidwarrington:


jackmcdade:
Fixed in ad7be2b (filename changed, couldn't merge). Thanks!

PR #184, opened 2020-08-25
Updated documentation to remove .app from dev-related TLD list

bradcis:
Now that .app is a register-able TLD, I'm guessing this is no longer exempt from the licensing fees.


jackmcdade:
Good point. Thanks!

PR #191, opened 2020-08-28
Fixed "throw ValidationException" code example

billmn:


jasonvarga:
Thanks but the withMessages method has new in it.  https://github.com/laravel/framework/blob/0b12ef19623c40e22eff91a4b48cb13b3b415b25/src/Illuminate/Validation/ValidationException.php#L71

PR #194, opened 2020-08-31
You also need doctrine/dbal for mysql

sytheveenje:
I'm using mysql and I also had to install doctrine/dbal.


jackmcdade:
Thanks!

PR #208, opened 2020-09-04
Fixed link to "Other Template Engines" page

danystad:
The "Other Template Engines" link under Additional Reading points to https://statamic.dev/other-template-engines instead of https://statamic.dev/template-engines which results in a visit to Clippy on the 404 page. He was less lonely for a minute.


jackmcdade:
Thank you!

PR #225, opened 2020-09-09
Clarify adding revision content to git repository

fitzage:
Just a small update to give the actual syntax for updating `.gitignore` to allow revision content in your repo.


jasonvarga:
Before we merge this, just putting a note to say that this change alone doesn't seem to work for me.

fitzage:
Hmm…might require a manual `git add storage/statamic/revisions`.

fitzage:
OK, when I’d tested with just the `!revisions` line, I had apparently added the revisions folder to the repo, which mean that the `!revisions/**` line worked. Turns out you need both of them, because `!revisions/**` only catches the children not the top level folder.

jasonvarga:
As we've seen first hand, messing with the gitignore is a bit confusing. We'd like to promote the better solution which would be to store your revisions somewhere more visible, like the content directory which you are likely already versioning.

The PR is appreciated though!

PR #239, opened 2020-09-17
Fix typo

jelleroorda:
That title speaks for itself, doesn't it 😄?


jackmcdade:
What an ironic typo lol

PR #241, opened 2020-09-17
Update bard syntax for sets and the views directory

### LINKED ISSUE #8: Change example of bard set partials to new antlers format
On this page: https://statamic.dev/fieldtypes/bard#templating

You should change the example that reads
```
partials/
|-- sets/
|   |-- gallery.html
|   |-- image.html
|   |-- poll.html
|   |-- text.html
|   |-- video.html
```
To:
```
resources/views/partials/
|-- sets/
|   |-- gallery.antlers.html
|   |-- image.antlers.html
|   |-- poll.antlers.html
|   |-- text.antlers.html
|   |-- video.antlers.html
```
This way people know where to put these partials and they'll be reminded of the new antlers naming for them.

jonassiewertsen:
The updated bard syntax is a little shorter and does look cleaner. 

To make it 100% clear where the partials can be stored, I did add the explicit directory. 

This PR closes #8

### DIFF
--- a/content/collections/fieldtypes/bard.md
+++ b/content/collections/fieldtypes/bard.md
@@ -104,13 +104,13 @@ Use tag pair syntax with `if/else` conditions to style each set accordingly.
 ```
 {{ bard_field }}
 
-  {{ if type == "text" }}
+  {{ if text }}
 
     <div class="text">
       {{ text }}
     </div>
 
-  {{ elseif type == "image" }}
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thanks! Like Jack said, we need to keep those Antlers bits. Instead of making you update the PR, which would result in a one liner, I figured you wouldn't mind if I just did it myself.

PR #245, opened 2020-09-19
Update modifiers.md

ryanAbbott:
This example broke for me while working through the docs. There is no modifier `uppercase` but there is `upper`. When changed locally my output matches the proposed output in the example.


jackmcdade:
Thanks for another uppercut!

PR #248, opened 2020-09-19
Update collections.md

ryanAbbott:
Remove closing parenthesis where open parenthesis is missing. Could always remove the commas surrounding ", intentionally or not," and replace them with parantheses too. Maybe toss in an exclamation mark for fun: "(intentionally or not!)".


jackmcdade:
:ok_hand:

PR #250, opened 2020-09-20
Add a note about outputting Bard image sets

JayGeorge:
I experienced some confusion trying to output a Bard image set until it was explained to me that it's not naturally augmented with Antlers unless you set a max value of 1. I hope this will prevent other newbies from running into this.

### DIFF
--- a/content/collections/fieldtypes/bard.md
+++ b/content/collections/fieldtypes/bard.md
@@ -139,6 +139,10 @@ resources/views/partials/sets/
 `-- video.antlers.html
 ```
 
+### Image Sets
+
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thanks but this is not specific to Bard. 

The information in this PR doesn't quite sound right anyway. It's never _not_ augmented. If you have max_files: 1, you get a single asset. Otherwise its a collection of assets. Either way, the same Antlers loop syntax can be used in both cases.

Can you explain what exactly you're doing and we'll be able to work out if maybe you're experiencing a bug, or where the docs needs to be clearer.

jasonvarga:
I added more information to the assets fieldtype docs in e77cbdbd51f8dbecaeb1ee95528b3714c4de4b6a.

https://statamic.dev/fieldtypes/assets#templating

JayGeorge:
Ah ok! In terms of what I was doing:

I was trying to output a Bard image set as per the exact example on the [Bard docs page](https://statamic.dev/fieldtypes/bard#content).

I think the docs need to specify that for the following documented code example to "just work" then there needs to be a max file limit of 1:

```
{{ elseif type == "image" }}

    <figure>
        <img src="{{ image }}" alt="{{ caption }}" />
        <figcaption>{{ caption }}</figcaption>
    </figure>

{{ /if }}
```

As someone new to Statamic I expected the docs example to work. It took quite a bit of back and for on Discord for someone to suggest adding a limit of 1 to the field for anything to output on the front-end.

PR #251, opened 2020-09-20
CP update vs composer update

curtismchale:
Make it clear that updating from CP means a terminal update will not do anything until you've changed how the statamic version is noted in composer.

### DIFF
--- a/content/collections/docs/installation.md
+++ b/content/collections/docs/installation.md
@@ -66,5 +66,15 @@ composer update statamic/cms --with-dependencies
 
 > You may prefer to run `composer update` to update _all_ of your dependencies listed in your composer.json file
 
+## Statamic is Not Updating Properly
+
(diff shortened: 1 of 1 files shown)


jasonvarga:
Failure has nothing to do with it. This is just how it works. If you upgrade within the CP we lock you to the specific version you choose/see.

curtismchale:
Okay, but when someone updates from the CP and it fails (because it will sometimes) and then goes to update from terminal it won’t update. I know that’s “just how it works” but lots of people won’t know that at all and will be left annoyed that the system won’t update. You’ll deal with support in Discord and tickets.

Feel free to revise the wording, but this isn’t common knowledge and should be documented. I had to ask in Discord today and once I was told...problem solved right away. I wouldn’t have had to ask if it was clear in the documentation on updating.

jasonvarga:
The docs have been updated.

PR #253, opened 2020-09-21
Explain how index / show works on collections.

### LINKED ISSUE #252: Explain how index / show works on collections
Some people on Discord have found that after scaffolding resources and setting a route for single entries their index URL will 404. However they expect it to work. 

The docs could do a better job of explaining that you need to either:

1) Add a page and link your index template.
2) Make a custom template route.

robdekort:
Attempt to close #252


jasonvarga:
Thanks but the intention would be to use the mount feature. We'll cook something up and put it there, that's a good spot.

robdekort:
I see. People would still need to know they have to add a page manually right? We'll see what you come up with :-)

jasonvarga:
How's that?

robdekort:
Yeah cool! That'll help :-) Nice!

PR #254, opened 2020-09-22
Added missing model parameter

chimit:
In the current default `config/auth.php` file there is no `model` parameter:

```php
    'providers' => [
        'users' => [
            'driver' => 'statamic',
        ],
    ],
```
So changing `statamic` to `eloquent` is not enough and causes a "ErrorException Undefined index: model in CreatesUserProviders.php line 82" error.


jasonvarga:
The current config/auth.php has all three variations, ready for you to just uncomment: https://github.com/statamic/statamic/blob/master/config/auth.php#L68-L82

Where are you looking?

chimit:
It was missing examples in the Cool Writings Starter Kit. Sent a PR there instead.

PR #261, opened 2020-09-28
Update antlers.md

MartinVandersteen:
Minimal copy change because of repetition. 
Not a native english speaker though so that might be intended, but I don't think so !


jasonvarga:
Perfect. Indeed looks like a brain fart. Not intentional 😄

PR #270, opened 2020-10-06
Document the new key param on the cache tag

damcclean:
This pull request adds the new `key` parameter to the documentation of the Cache tag.

This functionality was implemented in statamic/cms#2589.


jackmcdade:
Thanks Duncan!

PR #272, opened 2020-10-08
Update form example for dashboard widget

arimolzer:
The singular form example and the full example of possible widgets are inconsistent, and the singular form example, if attempted, will cause an error. Updated the form example to be inline with the full example.


jasonvarga:
Copy/paste error. Thanks!

PR #273, opened 2020-10-08
add config tag

edalzell:
You might want/need to "statamify" the language.

### DIFF
--- a/content/collections/tags/config.md
+++ b/content/collections/tags/config.md
@@ -1,0 +1,21 @@
+---
+title: Config
+description: 'Retrieves configuration data'
+intro: 'The config tag is used to get configuration values from your config files'
+stage: 5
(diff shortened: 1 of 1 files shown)


jesseleite:
Main problem with this is that it's not a tag.  It's just array data that's passed into your views, which you can see if you `{{ dump }}`.

edalzell:
This should be re-worded then?
image

jesseleite:
Yeah that sounds good, and maybe not in the `/tags` part of the docs, since it's technically not a tag.

edalzell:
Sorry @jesseleite, I'm confused with your comment. So not a tag? If so, there's not really anything to change?

Issue is that it doesn't show up in a search so people can't find it.

jasonvarga:
Pushed a commit and made some tweaks. It should become searchable soon once Algolia does its magic.

PR #275, opened 2020-10-08
Update fields.md

jimmyaldape:
Missing "to" on line 4

### DIFF
--- a/content/collections/docs/fields.md
+++ b/content/collections/docs/fields.md
@@ -1,7 +1,7 @@
 ---
 title: Fields
 template: page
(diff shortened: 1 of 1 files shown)


jimmyaldape:
To clarify, sentence should read `Fields are to be grouped into blueprints and fieldsets and can be reused in a number of different ways.`

However, it would also make sense to just remove "be" so that it reads, `Fields are grouped into blueprints and fieldsets and can be reused in a number of different ways.`

jasonvarga (changes requested):
Want to remove the `be` like you suggested instead? I think that reads better.

jimmyaldape:
You were right it reads better. Made the change.

PR #277, opened 2020-10-09
Changes to data retrieval

andershagbard:
- Make it more clear what an ID is.
- Add docs for findBySlug and findByUri


jasonvarga:
Thank you for pointing this out. I made some different changes based on your PR instead.

PR #279, opened 2020-10-13
Fix typo in 'Preving Stacks from closing' heading

paulstone:
Should be 'Preventing Stacks from closing'.


jasonvarga:
That's slang. It's how the cool kids say it. 😊

PR #284, opened 2020-10-19
Add a note around pointing web server at public folder

damcclean:
I've seen a few people struggle with getting Statamic setup locally and the issue was that they weren't pointing their web server at the `public` folder, instead they were pointing it at the root project folder.

Example: https://statamic.com/forum/4931-struggling-on-first-install

Hopefully adding a line like this will help to avoid any confusion.


jackmcdade:
Good one, thank you!

PR #289, opened 2020-10-22
Grammer change

cazepeda:
Add `be` to the sentence `You can choose which collections' entries will available to choose from.


jackmcdade:
Thanks!

PR #291, opened 2020-10-23
Fix {{ get_content }} example

damcclean:
The variable should be `related_by_id` as per the front matter in the example.


jackmcdade:
Thanks!

PR #292, opened 2020-10-25
Metrics are now used in a different way.

sandersjj:
The old way to use metrics are not implemented see `/statamic/cms/src/Forms/Form.php:201`. So in order to not confuse others I removed it from the docs.


jackmcdade:
Good call. We need to loop back around to these – it completely slipped off our radar.

PR #293, opened 2020-10-26
Add "hot tip" requisite information for cache tag.

jelleroorda:
Added a note/hot tip about making sure to use a caching strategy as a requisite for using the `cache` tag.


jasonvarga:
This is not true. Static caching is not a requirement for the cache tag. Please open an issue at statamic/cms that describes the problem. Thanks!

jelleroorda:
Just confirmed, false alarm 🙈.

PR #294, opened 2020-10-26
Update key for setting the time limit for half measure caching

jelleroorda:
The documentation for the expiry time for half measure caching seems to be incorrect (outdated?).

You can see this in the [ApplicationCacher.php](https://github.com/statamic/cms/blob/master/src/StaticCaching/Cachers/ApplicationCacher.php#L39), which inherits it from the parent here: [AbstractCacher.php](https://github.com/statamic/cms/blob/master/src/StaticCaching/Cachers/AbstractCacher.php#L58)

### DIFF
--- a/content/collections/docs/static-caching.md
+++ b/content/collections/docs/static-caching.md
@@ -130,9 +130,22 @@ A statically cached page will be served until it is invalidated. You have a seve
 
 ### Time Limit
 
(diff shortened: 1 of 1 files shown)


jasonvarga:
Oh that's a bummer.

Looks like I wrote the docs how I wanted it to word but never updated the code.

I'd say this is a bug, or something that should be updated in the code itself. I'd rather you define `expiry` than `default_cache_length`. There's no "default". It's just what it is.

jasonvarga:
Closing in favor of https://github.com/statamic/cms/issues/2744

PR #295, opened 2020-10-28
bard_field to bard

ItsMalikJones:
This change is minor and serves to help others get bard up and going quicker by using its default variable to render content 'bard' instead of 'bard_field'.


jasonvarga:
Thanks but we intentionally don't use the fieldtype handle as-is to show that it's the field name and not the fieldtype name.

PR #298, opened 2020-10-31
Typos fixed within Image Caching section

sjclark:
Fixed a few typos - looks like something borked as the formatting was also a little unusual!

> if there are a lot of **manipulation** on any given page **quest** etc


jackmcdade:
Looks good to me, thank you for taking the time!

PR #302, opened 2020-11-05
Remove --with-dependencies parameter from installation docs

damcclean:
This pull request removes the `--with-dependencies` parameter from the update documentation. As it would throw errors if some people are running Laravel 7 sites. I'm sure this parameter was added for a reason...

See statamic/cms#2803.


jasonvarga:
We want the dependencies to be updated too.

PR #303, opened 2020-11-11
Update git-integration.md

robdekort:
Add examples for Forge and Ploi on how to skip deployment when using git integration and when you use a custom commit message.

### DIFF
--- a/content/collections/docs/git-integration.md
+++ b/content/collections/docs/git-integration.md
@@ -177,6 +177,26 @@ For example, you can append `[BOT]` to the commit message so that you can select
 ],
 ```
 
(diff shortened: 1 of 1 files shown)


jackmcdade:
I've never heard of Ploi, interesting. Would you be able to add links to right services so the context won't be missed?

jackmcdade:
Great, thank you!

robdekort:
Done! Yeah, Ploi looks great. We've been talking about it in the server channel often on Discord. Some people use it and really prefer it over Forge. I'd like to switch but it's a lot of work.

robdekort:
Welcome! It's cheaper, has more features (something like envoyer bascially included) and development and support is very active.

Issue #305, opened 2020-11-12
Broken link to fieldsets docs


CasperEngl:
## Issue

On the page https://statamic.dev/fields#blueprints-amp-fieldsets under "Blueprints & Fieldsets" the link to Fieldsets currently goes 404.

## Suggested change

Change the link from: https://statamic.dev/fieldsets
To this link: https://statamic.dev/fieldtypes/fieldset

## Bonus

[Ohdear.app](https://ohdear.app/) has a broken link crawler. Seeing as the /fieldsets page sends the 404 status code, ohdear can find such links and tell you about them.

janparkio:
I like the suggested change, however, the link: https://statamic.dev/fieldtypes/fieldset feels almost pointless.
Compared to other pages in the Statamic documentation, this one feels very much lacking.

The methodology to use Fieldsets is beautifully simple, but by no means intuitive for a beginner, and it is actually a really great tool, I am surprised nobody changed it yet.

As you can see I've made a similar issue #296 and at the moment is closed. But the issue was not solved.

PR #306, opened 2020-11-12
Fix broken link to fieldsets

CasperEngl:


damcclean:
Closes #305.

jackmcdade:
Ah, you know what – the fieldsets page is actually missing. Linking to the fieldset fieldtype isn't the bet explanation of what they are and how they're different from Blueprints.

PR #307, opened 2020-11-12
KB article for PHP extension requirement gotcha

jmalko:
Worked through an elusive issue in Assets and Glide section where Exif extension was preventing image manipulation to work.  Error was difficult to track down without getting creative.


jackmcdade:
Nice one, thank you! Can you remove the `updated_at` and `updated_by` fields? I'd love to merge it and that'll keep it cleaner. :)

jmalko:
Got it.  Thanks Jack.

jackmcdade:
👏

PR #309, opened 2020-11-13
Added instructions to "Create an About page"

mikeritter:
I really wanted to dive into each tool in the workflow,  but kept it simple.

Don't know whether to write in first or second person.


ebeauchamps (commented):
typo: "now let's secect "default"" should be "now let's select "default""

PR #311, opened 2020-11-15
Make it clear to add `antlers` suffix to replicator partials

klickreflex:
I'm new to statamic and had issues following along these docs because my variables weren't rendered properly when moving replicator sets into partials as suggested. I've found that it works as soon as I add the antlers suffix to my partials, so I guess it would be nice having this in the docs.


jasonvarga:
Thanks!

PR #313, opened 2020-11-19
Update collection.md

freshface:
I think this is better.
The previous text is from v2?


jackmcdade:
I'm good what that, thank you!

PR #314, opened 2020-11-19
Update principles.md - Typo

cnc137:
Removed "a" in "Statamic operates in flat file “mode” by default, which reduces complexity compared to other architectures, and opens up a many possibilities like:".


jackmcdade:
Thanks!

PR #326, opened 2020-12-10
Adds info about restarting your server

HellPat:
If you have statamic running via `php artisan serve` and change the `.env`-File the server seems to have the old filed cached.
Restarting the server helps a lot :-)


jasonvarga:
Changing .env variables works without restarting your server. Are you caching your config? Don't do that locally.

PR #337, opened 2020-12-14
Change `filter` to `visibleTo`

damcclean:
In the documentation about [creating a custom action](https://statamic.dev/extending/actions#filtering-actions), it references an old `filter` method which has since been renamed to `visibleTo`. This PR just fixes that.

It was pointed out in [this Forum post](https://statamic.com/forum/5067-filter-method-on-custom-actions-cp)


jasonvarga:
Thanks!

PR #338, opened 2020-12-16
Fix file name and array for sort parameters in locales tag docs

twd3:
I got confused while reading the docs for the locales tag. It still had the V2 system.yaml and the locales array referenced un der the sort parameter info. Changed to sites.php and sites array under the sort parameter...


jasonvarga:
Thanks!

PR #339, opened 2020-12-17
Added starter kits from CLI

makkinga:
I've added the missing starter kits that are featured in the Statamic CLI


jasonvarga:
Thanks!

PR #342, opened 2020-12-18
`results` is not needed

edalzell:
If you use the examples as listed they don't work because there is no `results`.


jasonvarga:
They're needed on the entries example because of pagination.

PR #348, opened 2020-12-22
Removed 'new' keyword

St3Ko:
The static 'withMessages' method is not a constructor, so this code example will result in a ParseError: "syntax error, unexpected 'withMessages' (T_STRING), expecting variable (T_VARIABLE) or '$'"


jasonvarga:
Thanks!

PR #352, opened 2021-01-01
Document where to configure language(s) used by the `iso_format` modifier

miclf:
It took me a while to figure this out, because I was wondering why date translations were not taking my `locale` into account… So I thought it would be nice if I could avoid some frustrations to other beginners like me.

Feel free to change the wording, but I think it would be quite helpful to explicitly say that the language has to be configured in the `config/statamic/sites.php` config file. Being experienced with Laravel, it totally made sense to me to expect the framework’s `locale` setting to be taken into account by Statamic’s `iso_format` modifier, and the docs said nothing to prevent me from going in this wrong direction.


jackmcdade:
Thank you!

PR #353, opened 2021-01-03
Correcting hash anchor in Glide tag and Variables link in table of contents

### LINKED ISSUE #278: Wrong hash anchor in Glide tag
On the page https://statamic.dev/tags/glide the link "See the Focal Crop section for more details." has the wrong hash id so nothing happens when you press it.

Instead of linking to #focal-crop it should be #focal-point-cropping

nmusayev:
Fixes #278


jasonvarga:
Thanks!

Issue #356, opened 2021-01-10
Undocumented value_header and key_header variables for dynamic arrays 


sjclark:
https://github.com/statamic/docs/blob/master/content/collections/fieldtypes/array.md
Add undocumented `value_header` and `key_header` variables for dynamic arrays and/or add to Control Panel.
```
value_header: Type of Bacon
key_header: Why is it awesome?
```
<img width="383" alt="Screen Shot 2021-01-10 at 2 13 36 pm" src="https://user-images.githubusercontent.com/4531017/104113460-13d1c080-534e-11eb-806d-16005a1701c6.png">

Also important as this has changed from the previous `text_header`?

PR #357, opened 2021-01-10
Documented value_header and key_header variables for dynamic arrays

sjclark:
Add undocumented `value_header` and `key_header` variables for dynamic arrays and/or add to Control Panel. 
As per: https://github.com/statamic/docs/issues/356


sjclark:
Also added `add_button` variable

jackmcdade:
Nice one, thank you!

PR #362, opened 2021-01-18
Add deployment bash script example for cleavr

armgitaar:
Added an example bash script for Cleavr to prevent auto-deployment in same vein as the Forge and Ploi examples.


jasonvarga:
Neat, thanks!

PR #365, opened 2021-01-21
Important note on cached config

jakubjo:
When you install Statamic into an existing Laravel app, you might break your production site if you do not delete the cached config before the first Statamic installation.


jackmcdade:
Good one! Thanks!

PR #367, opened 2021-01-25
update oauth.md

aosinski93:
change callback description from 'withUser' to 'withUserData'


jasonvarga:
This is not accurate. Please open an issue on statamic/cms with the problem you're running into. Thanks!

PR #368, opened 2021-01-25
Add routing tip for new collections

friartuck6000:
Adds a tip pointing out that Statamic doesn't use a default route rule for new collections.

This is probably obvious to most people—but for some who are coming from different ecosystems, or who sometimes forget their common sense (like me), it might help save a few minutes of confusion.


jackmcdade:
This page is better for your addition. Thank you!

PR #369, opened 2021-01-25
Should it be PreventRequestsDuringMaintenance.php?

helloDanuk:


jasonvarga:
Yup, it changed in Laravel 8. Nice one.

PR #370, opened 2021-01-27
Update widgets.md

bomshteyn:
The forms widget loops over `fields` (`@foreach($fields as $key => $field)`). Yet when using the current example you only see the date submitted column.

I just wasted half an hour resaving my yaml files and other settings thinking that it was some issue related to that. To discover that in order to display the fields you need to specify them in the config. I was under impression that the listable param would be used for this.


jackmcdade:
Great addition, thank you!

PR #373, opened 2021-01-29
Upda to php artisan statamic:multisite command

freshface:
Clearing the cache BEFORE running php artisan statamic:multisite is recommanded.


jasonvarga:
The command already does that now. Update statamic.

PR #384, opened 2021-02-08
Add example for sorting entries in collection widget

sebszocinski:
Update docs since this feature was added: https://github.com/statamic/cms/pull/1343


jackmcdade:
Thanks!

PR #385, opened 2021-02-10
Add Multi-Site section to Content API

fjahn:


fjahn:
Just noticed that you can also use `filter[locale]=default` instead of `filter[site]=default`. I still think that `site` is the better choice here for the sake of readability (the terms 'locale' and 'site' are used interchangably in the code, but not in the docs).

However, I am not sure which of the two is more stable.

PR #391, opened 2021-02-12
Update sites.md

rrotteveel:
To loop through the sites, the antlers (tag) should be looping over sites and not site. Small thingy, probably a typo.


jackmcdade:
Thanks!

PR #392, opened 2021-02-15
Update form-create.md

robdekort:
Explain how to dynamically render a selected form via the Form Fieldtype. Question comes up often.


jackmcdade:
Have you tried `{{ form:create :in="form_fieldtype:handle" }}`? If that works (it should), would you mind updating it? Interpolation over vars-in-params is our preferred syntax. 😊

robdekort:
Yes that is nicer. I'll check right away.

robdekort:
Doesn't work.

jackmcdade:
:poop:

PR #399, opened 2021-02-21
Add docs about how to get started with front-end frameworks.

jelleroorda:
@drikishuk and I were talking on Discord about how we were having some trouble before when we just started out with Statamic, so we've put together some starting docs to hopefully save other developers a few hours.


jackmcdade:
Very nice, thanks fellas! ❤️

PR #404, opened 2021-03-04
Update static-caching.md

Liopia:
Quick fix on a double faster faster. (Unless it's so fast that the double faster is the only way to properly express its faster-ness?)


jackmcdade:
I was kinda thingking "fast, faster, fastest" for a second there but decided to merge your PR. Thank you!

PR #405, opened 2021-03-04
Update list.md

preeco-privacy:
field handle is defined as product_ideas so the loops need to use this handle


jasonvarga:
Thanks. The example on that page used to be a todo list and must have missed those when updating.

PR #408, opened 2021-03-09
fixed method; added csrftoken; added submit button

j3ll3yfi5h:
replaced "action" by "method", added csrftoken field and submit button


jackmcdade:
Thanks!

PR #413, opened 2021-03-19
Update nav.md

klickreflex:
Document `is_current` property

The only way to find out that `is_current` can be used to easily modify navigation behavior for external links (i.e. adding `target="_blank"`) was through dumping variables. This should be documented imho.


jackmcdade:
Thank you!

PR #423, opened 2021-03-26
Added more tags to modifiers

adnweedon:
Added some missing tags to modifiers that say they apply to multiple categories


jasonvarga:
Thanks!

PR #425, opened 2021-03-26
Independent Auth Guards

duncanmcclean:
This PR turns [Jason's comment](https://github.com/statamic/cms/pull/3143#issuecomment-793101991) on my 'Independent Auth Guards' PR into a knowledge base article, which I've linked next to the other user related links (probably where I'd find myself if I was looking for such thing).

Related to statamic/cms3143

### DIFF
--- a/content/collections/docs/users.md
+++ b/content/collections/docs/users.md
@@ -1,14 +1,14 @@
 ---
 title: Users
-intro: 'Users are the member accounts to your site or application. What a user can do with their account is up to you. They could have limited or full access to the Control Panel, a login-only area of the front-end, or even something more custom by tapping into Laravel.'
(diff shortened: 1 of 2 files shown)


jasonvarga:
Thanks Duncan!

connor11528:
@duncanmcclean thank you for updating the documentation :) In the example shared would it make sense to have the Statamic config reference the `statamic` guard? In the current example the statamic config references the web guard. 

Thinking to have it reference the auth guard we created it'd need to be something like this?
```
// config/statamic/users.php

'guards' => [
    'cp' => 'statamic', // the guard when using the cp
    'web' => 'statamic', // the guard when using Statamic frontend routes
],
```

jasonvarga:
Feel free to make a PR to the docs.

PR #433, opened 2021-04-06
Explain availability of is_entry and is_term variables

arthurperton:


arthurperton:
Sorry, I have to double check first..

arthurperton:
There is more going on here. Seems there is a bug in de Search tag when `supplement_data` is `false`. Better report / fix that instead of updating the docs.

PR #437, opened 2021-04-15
Nav:breadcrumbs tag: Clarify the meaning of the include_home parameter

acerempel:
Currently the docs for the `nav:breadcrumbs` tag claim that the `include_home` parameter *removes* the home page. That doesn't make sense: it *adds* the home page to the beginning of the breadcrumbs. This PR changes the description of the parameter to reflect this.


jackmcdade:
Nice catch, thansk!

PR #441, opened 2021-04-16
Clarify other fields can be used for template variable

### LINKED ISSUE #262: Explain how to set default templates for blueprints
This question comes by a lot.

Use a hidden field called template with a default value of the desired template name.

JonKaric:
Doesn't directly fix #262, but adds clarification that other fieldtypes (including hidden) can be used too.


jackmcdade:
👍

Issue #281, opened 2020-10-13
Broken Link & No Documentation on Roles


conradfuhrman:
On the permissions page https://statamic.dev/extending/permissions the link on the top of the page in the first sentence Roles links to a 404 page.

PR #442, opened 2021-04-16
Fix broken role links

JonKaric:
Fixes broken role links #281. Kinda a half fix as there's no docs for roles currently


jackmcdade:
<3

PR #446, opened 2021-04-19
Add an extra introduction on how to get started extending Bard.

### LINKED ISSUE #424: Extending Bard docs
This whole page is unfortunately very unclear to me: https://statamic.dev/extending/bard#extensions

In my opinion it would benefit from the following:

- What is TipTap (in relation to Statamic and Bard)
- What is ProseMirror (in relation to Statamic and Bard)
- What is Mark
- Where do you put this JS Code
- What is ProseMirror rendering
- TipTap library / API?
- Example: simple button that does a simple thing like adding a class or wrapping in stuff in an element
- Less assumed knowledge

jelleroorda:
Closes #424.

### DIFF
--- a/content/collections/extending-docs/bard.md
+++ b/content/collections/extending-docs/bard.md
@@ -1,12 +1,26 @@
 ---
 title: Bard
 stage: 1
+intro: 'The Bard editor used in the control panel is based on [TipTap](https://tiptap.scrumpy.io/) (which is a Vue WYWIWYG editor), which in turn is a wrapper around [ProseMirror](https://prosemirror.net/docs/guide/) (which is a plain JavaScript WYSIWYG editor).'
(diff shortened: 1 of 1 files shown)


jelleroorda:
@robdekort Can you check if you're still missing things, whether things are unclear, or would need any further pointers for?

robdekort:
Nice, I think this really really helps a lot. It's already way clearer to me. Although I'm fairly positive now I'm def. not gonna burn my hands on extending Bard, which is based on TipTap, which is a wrapper around ProseMirror 🤯

😉

jelleroorda:
Well it’s definitely something that will cost you a lot of time, and is also pretty prone to bugs if you ask me... so that’s probably a wise choice 😉

jackmcdade:
Thanks for the fellas! Made a few little tweaks but this is super helpful. Much appreciated!

PR #447, opened 2021-04-19
Extend the nav:breadcrumbs documentation with a multisite caveat.

### LINKED ISSUE #436: Improve the nav:breadcrumbs tag
It needs some explanation on how it works out what each crumb should be.
When it would skip segments, etc.

See https://github.com/statamic/cms/issues/3523

jelleroorda:
I've added some extra documentation for the `nav:breadcrumbs` tag, as I understand it explained [by Jason](https://github.com/statamic/cms/issues/3523#issuecomment-819146401).

Closes #436


jasonvarga:
I think we can reword this. It's not a multisite specific caveat. It's how it works in general.

PR #449, opened 2021-04-20
Update oauth.md. Missing service provider when adding additional socialite providers.

matiit:
Add a missing step needed for the SocialiteWasCalled event to be ever called.

### DIFF
--- a/content/collections/docs/oauth.md
+++ b/content/collections/docs/oauth.md
@@ -108,6 +108,16 @@ Require the appropriate provider using Composer:
 composer require socialiteproviders/dropbox
 ```
 
+Ensure the `SocialiteProviders\Manager\ServiceProvider` is present in `config/app.php`:
+
+```php
+'providers' => [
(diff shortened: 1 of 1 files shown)


jasonvarga:
This shouldn't be necessary. What Laravel version are you using?

matiit:
> This shouldn't be necessary. What Laravel version are you using?

Thanks for getting back to me.

```json
"name": "laravel/framework",
"version": "v8.37.0",
```

```json
"name": "statamic/cms",
"version": "v3.1.5",
```

```json
"name": "laravel/socialite",
"version": "v5.2.3",
```

```json
"name": "socialiteproviders/manager",
"version": "4.0.1",
```

It seems that without specifying the `SocialiteProviders\Manager\ServiceProvider` in _app.php_ the listener for `SocialiteWasCalled` event is not being called and the custom socialite driver/provider is not ending up being registered.

PR #450, opened 2021-04-21
Update nav.md

Eworm:
Adding an example of how to get and show the children of the current page.


jackmcdade (approved):
Thanks — nice addition!

PR #451, opened 2021-04-21
Adding missing braces

Eworm:
The uri has to be wrapped in braces otherwise it doesn't work.

### DIFF
--- a/content/collections/tags/nav.md
+++ b/content/collections/tags/nav.md
@@ -156,7 +156,7 @@ Use the `uri` to get the children of the current page.
 
 ```
 <ul>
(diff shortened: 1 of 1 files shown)


jackmcdade:
Not true. That prefixed `:` tells it to interpolate the name of the variable so you don't need to parse the contents with braces.

https://statamic.dev/antlers#variables-inside-tag-parameters

Eworm:
Without the braces the nav gets the siblings of the current page, not the children.

Eworm:
Testing it in a site I'm working on as we speak 🙂

PR #456, opened 2021-04-30
Added an exemple about displaying each locale excepted the current one

FR6:


jackmcdade:
Thanks for the suggestion! I reworked the page to make everything a bit more clear and made sure to cover this example (in a simpler way).

PR #458, opened 2021-05-03
Fixing typo : secect -> select

OlivierFortier:
Fixed a typo encountered while working through the quick start.

> now let's _secect_ "default" for our About page. 

should be 

> now let's _select_ "default" for our About page.


jackmcdade:
Thank you!

PR #461, opened 2021-05-05
Replace relate tag reference with get_content

robdekort:
Since the relate tag doesn't exist anymore in v3.


jackmcdade:
Nice catch, thank you!

PR #467, opened 2021-05-10
Move "Protected Page" Screenshot

cnc137:
Move "The default password protected login screen." screenshot to be in the "Password" section instead of "Authentication".


jackmcdade:
Good one!

PR #470, opened 2021-05-15
Command/Ctrl+K keybind that focuses on search input

ncla:
Switching between Tailwind and Statamic docs I found it frustrating that there was no CTRL+K binding to focus on the search input.

Borrowed/inspired from how Tailwind does it in their docs. https://github.com/tailwindlabs/tailwindcss.com/blob/7617a606ee89065144bcfe3e6b35d2938e707c0a/src/components/Search.js

Moved the Alpine x-data outside to <script> tag because I could not figure out how to dynamically create a keybind.

Unfortunately can't test if it works on MacOS as intended.


jackmcdade:
We already have a `/` to focus search, is that not good enough?

ncla:
I think it depends who you ask. Some apps (e.g. Discord) use CTRL+K, some sites use the slash button. And as Tailwind is pretty relevant/close to Statamic in the "eco-system", having parity with other documentation sites would be nice to have in my opinion.

If not, I'll just bust out a hacky user script for myself. Your call, boss! :smile:

jackmcdade:
If this helps you, then I'm fine with it :)

PR #472, opened 2021-05-20
Use less generic set types to make Bard looping example clearer

ncla:
It may seem insignificant but I think this can be beneficial for those people (like me) who scour the docs by looking at the example codes instead of reading the whole page. The set examples like "text" and "image" are in my opinion too generic that it can lead you to think that you can loop through normal rich text editor nodes, which you cannot currently.

PR is inspired by this situation https://discord.com/channels/489818810157891584/489819906540568593/844982271605866536

Let me know what you think.

### DIFF
--- a/content/collections/fieldtypes/bard.md
+++ b/content/collections/fieldtypes/bard.md
@@ -104,18 +104,22 @@ Use tag pair syntax with `if/else` conditions to style each set accordingly.
 ```
 {{ bard_field }}
 
-  {{ if type == "text" }}
+  {{ if type == "poll" }}
 
-    <div class="text">
-      {{ text }}
-    </div>
-
-  {{ elseif type == "image" }}
-
-    <figure>
-      <img src="{{ image }}" alt="{{ caption }}" />
-      <figcaption>{{ caption }}</figcaption>
+    <figure class="poll">
+      <figcaption>{{ question }}</figcaption>
+      <ul>
+        {{ options }}
+          <li>{{ text }}</li>
+        {{ /options }}
+      </ul>
     </figure>
+
+  {{ elseif type == "hero_image" }}
+
+    <div class="heroimage">
+      <img src="{{ hero_image }}" alt="{{ caption }}" />
+    </div>
 
   {{ /if }}
 
@@ -133,9 +137,8 @@ An alternative approach (for those who like DRY or small templates) is to create
 ``` files
 resources/views/partials/sets/
 |-- gallery.antlers.html
-|-- image.antlers.html
+|-- hero_image.antlers.html
 |-- poll.antlers.html
-|-- text.antlers.html
 `-- video.antlers.html
 ```
 


jackmcdade:
This is great. I’ve been doing this kind of thing whenever I touch a section of the docs, so thanks for jumping ahead!

jasonvarga:
Great idea indeed, but now there's no text.

jackmcdade:
Great point @jasonvarga! It's back.

ncla:
Didn't expect to be merged this fast. Thanks! But I am wondering why text set was put back, in my opinion it's now two steps forward, one backward. The removal of text set was intentional.

duncanmcclean:
By the sounds of things... I think it caused a bug on the docs site so they had to roll it back.

jackmcdade:
Text isn’t the name of a set, it’s the name of the text in between sets. If you don’t show it, people will never know how to render the content they write.

ncla:
Ah, I understood now. I thought it was not possible as it was either you use single tag for rich text only or tag pairs if you have sets only, not both. I'll create another PR to make this clearer if you don't mind.

jasonvarga:
Bard is always going to have text. Like Jack said, the text is the bits between the sets.

If you want a field that has only sets and no text, you should use a Replicator instead.

If you want to make that clearer, go for it!

montemartin:
This change is helpful for understanding what sets are for and how they work, but I doesn't completely solve the initial misunderstanding, as evidenced by the need for additional explanation. To truly clarify the point in the docs at least some of this conversation needs to be included explicitly. There are two absolute statements that are still only inferred by the language used. 

First, in Discord @jasonvarga said "If your bard field doesn't have any sets, you **have to** use the single tag". The key words being "Have to". There is no way to know what @ncla was trying to achieve initially: `{{ bard_field }} {{ dump }} {{ /bard_field }}` wouldn't work. I had also tried the same thing, because it seems like it should work.

Second is the comments above regarding `type=="text"`. It still isn't clear from the code that the text field is necessary. Yes, it's evident once you dig through the actual output from a bard field, but that takes significant extra work to understand the concept. 

@ncla Here's some rough ideas in case you do get to another PR. I also think some well place code comments would help.

**Without Sets**
If you are only using Bard as a rich text editor ~~and have no need for sets you would~~ without any sets, use a single tag to render the content. 

**With Sets**
Use tag pair syntax with if/else conditions to style each set accordingly. Be sure to include a conditional for `type=="text"`. It will handle content that is entered outside of  other sets.
```
{{ bard_field }}

  {{ if type == "text" }} {{# <- Don't forget this! #}}
    <div class="text">
      {{ text }}
    </div>
...
```

ncla:
Yep, I was thinking something like this. In addition, remove the "text" set type from the file list below as well, because it is not a set type as the directory name and template suggests. It's just simply.. text.
image

montemartin:
I think in that case you'd need to include an if statement around the partial along with a text handler. Something has to handle those type==text cases.

PR #475, opened 2021-05-22
Document Bard text type, recommend using Tailwind typography plugin

ncla:
Continuing from here: https://github.com/statamic/docs/pull/472

1. Documented the `text` type so it is clearer that it is the content between the sets and not another set.
2. Guide the developer in right direction on how to style rich text editor content by recommending Tailwind typography plugin.  I hope this reduces instances where a developer might think about styling each individual node. While Erin did point out that it is possible to style each node, it seems like an approach that should not be recommended in the docs.

### DIFF
--- a/content/collections/fieldtypes/bard.md
+++ b/content/collections/fieldtypes/bard.md
@@ -97,9 +97,11 @@ If you are using Bard just as a rich text editor and have no need for sets you w
 {{ bard_field }}
 ```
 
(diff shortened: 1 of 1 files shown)


jackmcdade:
Thanks for this PR! I actually really dislike the Tailwind Typography plugin because it's overwhelmingly complex to explain and customize, but I do appreciate it. I reworded the bit about the text type based on your feedback and pushed that. Appreciate the help and nudge!

PR #491, opened 2021-06-11
fix oauth callback method name

dmgawel:
The paragraph mentions `withUser` method name, while in fact the `withUserData` is both a valid name and the one used in an example.


jasonvarga:
Thanks!

PR #493, opened 2021-06-18
Update bard docs for custom extensions

jonassiewertsen:
It seems necessary to have all listed methods, even if returning an empty array.

Not having those did throw errors on my end, which I couldn't connect to the real problem. Adding this in the docs would make the dev experience much nicer and would have saved me some hours. 

Not making those methods required in the core would be a different approach.


jasonvarga:
Thanks Jonas. We could make those optional eventually, but this will help in the mean time.

jonassiewertsen:
Thanks for the fast merge!

PR #496, opened 2021-06-23
Taxonomy entries count

mikemartin:
To sort by most popular terms the sort must be entries_count descending.


jackmcdade:
Thanks!

PR #498, opened 2021-06-24
Glide: Extend description for quality property

fjahn:
Add webp to the list of formats which are affected by quality.

This is not documented in [the glide docs](https://glide.thephpleague.com/2.0/api/encode/#quality-q), but I just tested it, and `webp` is also affected by `quality`.


jackmcdade:
Neat!

PR #503, opened 2021-07-12
fix template tag

j3ll3yfi5h:
https://discord.com/channels/489818810157891584/489819906540568593/864083857791385622


jackmcdade:
Believe it it or not, it's not. `{{ template }}` is the _name_ of the template, and `{{ current_template }}` is the special variable were we render the template _contents_. Thanks though!

jackmcdade:
Sorry, I read `template_contents` in my head there. Yes you're totally right 😂

vannut:
I broke my head on this one.

PR #505, opened 2021-07-13
Clarify save_html option for bard fields

fitzage:
In retrospect it should have been obvious that I couldn't have bard fields with sets save as HTML, but it would be a good idea to have it in the docs.


jackmcdade:
Thanks @fitzage!

PR #512, opened 2021-07-22
Remove the butik starter kit

jonassiewertsen:


jackmcdade:
Happy to merge, curious as to why though!

PR #521, opened 2021-08-15
Update quick-start.md

eugenevdm:
Add `views` below resources as it's the correct root for the partial


jackmcdade:
Nice catch, thank you!

PR #529, opened 2021-09-02
Update taxonomies.md

martinhummer:
{{ results }} tag is not necessary here.

### DIFF
--- a/content/collections/docs/taxonomies.md
+++ b/content/collections/docs/taxonomies.md
@@ -143,9 +143,7 @@ When on a [term route](#routing), you can list the entries by using an `entries`
 ```
 {{ entries paginate="5" }}
   <ul>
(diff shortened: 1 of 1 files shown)


jackmcdade:
Nice catch – thanks!

jasonvarga:
It is necessary though. It's using pagination.

martinhummer:
You are right Jason.

I stumbled over this, because I used that snippet without the "paginate" argument, like this:

```
{{ entries }}
            {{ results }}
                <ul>
                    <li><a href="{{ url }}">{{ title }}</a></li>
                </ul>
            {{ /results }}
 {{ /entries }}

```
{{ results }} is not present in that case, and it took me quit a while to figure out that i've to remove {{ results }} to make it work.

PR #531, opened 2021-09-03
Correct a typo on the Collections page

BrianLangevin:
The word "respect" was misspelled as "repsect"


jackmcdade:
Thanks!

respect

PR #532, opened 2021-09-03
Typo: Use correct url for composer website

adamlindqvist:
http://composer.org/ is not valid. It should be https://getcomposer.org/


jackmcdade:
Thank you! What a rookie mistake – always double-check URLs jack!

PR #535, opened 2021-09-11
Use npm ci instead

jonassiewertsen:
I would suggest using `npm ci` instead of `npm install`:

https://dev.to/visuellverstehen/why-to-use-npm-ci-instead-of-npm-install-mfd


jonassiewertsen:
Not sure why it's marking line 167 as changed 🤷‍♂️

PR #558, opened 2021-10-17
Adds config/shceme info to example.

ChristianPavilonis:
The previous example does not include how to add additional config/scheme info to custom drivers. This stumped me for a time while building my own.


ChristianPavilonis:
It does not work as documented. But when working with the example I don't have access to `$config` or `$scheme` properties in the protector class. I'd think that it would call the `protector()` method on the `ProtectorManager` class.
I can, however, do `return $this->protector(new CoinFlip, $scheme, $config);` within the closure, because the Manager class binds the context to its self in the closure.
Should I open an issue on the cms repo?

jasonvarga:
Yes, open an issue. That's what I meant.

PR #562, opened 2021-10-21
findBySlug() requires additional parameter

leighgibbo:
I know it's deprecated, but the `findBySlug()` method also requires a collection-handle parameter to be passed along with the slug.


jasonvarga:
Thanks! I'd rather it just not be there at all.

PR #578, opened 2021-11-04
where clauses with collection do not need an operator

j3ll3yfi5h:


jackmcdade:
That is true, but that block right there is explaining how it works so we want to show the full method declaration and then teach the shorthand. You'll notice we don't use 3 argument blocks anywhere else unless necessary. :) Thanks though!

j3ll3yfi5h:
@jackmcdade but it did not work like this in my current project, I still think its wrong. Am I wrong?

jackmcdade:
What do you mean? What did you try specifically and what error did you receive specifically?

jasonvarga:
Yeah it doesn't work if you add the operator. Specifically for `where('collection'`

PR #591, opened 2021-11-10
Update collections.md

marflow:


marflow:
sorry! – read `blog | things` but `listings|piped`is correct. Late here, I´m stopping it now ;)

PR #593, opened 2021-11-11
Add store as reserved field word

robdekort:
It only took me three hours to find this out...

### DIFF
--- a/content/collections/tips/reserved-words.md
+++ b/content/collections/tips/reserved-words.md
@@ -17,6 +17,7 @@ This is the list of reserved words you shouldn't use as field names, in addition
 - `if`
 - `reference`
 - `resource`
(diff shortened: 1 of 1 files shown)


jackmcdade:
If this is true it should be added to the FieldsController as well: <https://github.com/statamic/cms/blob/3.2/src/Http/Controllers/CP/Fields/FieldsController.php#L71-L82>

jasonvarga:
I don't see why it should be reserved. Maybe we can fix the underlying problem.

robdekort:
You can check my support mail, it has a zip attached.

robdekort:
Reach out when you have q's of course!

jasonvarga:
Closing this, we'll sort it out over support.

PR #594, opened 2021-11-11
group_by example

robdekort:
Two collections:
* Blog
* Author

This will fetch you a list of authors that wrote a blog post.

### DIFF
--- a/content/collections/tags/collection.md
+++ b/content/collections/tags/collection.md
@@ -407,7 +407,7 @@ For example, if you want to group some dated entries by month/year, you could do
 {{ /collection:articles }}
 ```
 
-Another example, lets say you have a `blog` and an `author` collection. The blog collection has an author field and you want to get a list of authors that wrote a blog post:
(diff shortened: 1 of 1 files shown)


jackmcdade:
Hmm, not sure it's described quite right. Do you mean blog collection and author _field_? Also, an output might help users grok the example better.

robdekort:
Does this help? It came up in discord and Erin and I found it a nice solution as it's native antlers. Figured it would be nice in the docs.

jackmcdade:
I don't see any interaction with the authors collection in this example though – am i misunderstanding what you're doing? I'm sorry I'm so confused here.

robdekort:
Haha that's ok. Means my PR is not clear right? I'll add to it. Def. interacting with authors. See the group_by attribute.

robdekort:
Ok one more try because I really like you Jack....

jackmcdade:
I think this confuses me because I wouldn't use an Author collection, i'd just use the users for that. Plus, `author` is a special field name that enables the per-user permissions inside Collections. 🤔

PR #595, opened 2021-11-11
Link directly to Streamline Light page

jacksleight:
Streamline offer lots of different icon packs now, but the link just goes to the home page. I'm 99% it's Streamline Light, right?


jackmcdade:
About 90% of the icons are Light, yeah. Nice one, thanks Jack!

PR #601, opened 2021-11-14
Adding a custom selection color

haudraufhaun:
With this PR, just the background color of selected text get's changed. I just thought it could make the page even more fancier, but if you got any objections that would be fine as well. 

I'm not a designer guy, so when you got an better idea in terms of coloring, please tell :)

## screenshot
demonstration


jackmcdade:
I like it! Thank you!

PR #602, opened 2021-11-15
Update local.md

pauldoerwald:
The setup instructions don't include any steps for installing Laravel Valet, and having Valet is key to getting Statamic to run, at least on macOS.

### DIFF
--- a/content/collections/docs/local.md
+++ b/content/collections/docs/local.md
@@ -21,6 +21,7 @@ To install Statamic locally you will need the following:
 - A computer running MacOS, Windows, or Linux
 - A supported version of [PHP](https://php.net) (we recommend PHP 8)
 - [Composer](https://getcomposer.org) to manage PHP packages
+- [Laravel Valet](https://laravel.com/docs/8.x/valet) if running on macOS to host the site
 
 ## Install Statamic CLI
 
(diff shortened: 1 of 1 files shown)


jasonvarga:
We aren't really assuming you're using Valet though. 🤔  You could be using any type of server. Not even Mac specific.

pauldoerwald:
Fair. The instructions don't say ANYTHING about using a server. They say that after you install Statamic, you can access the site at http://$project_name.test, but you actually can't unless you've set up a server. I was misled on Chrome/macOS because it responded with "403 Forbidden (Apache Server at clockk.test)". That, to me, suggested that there was a server running (and there is, somewhere...).

Instead of my words, perhaps there should be a step that says "Set up a server here that will answer for your site. You could use Valet". But the lack of a step implies that Statamic will be running, which it isn't.

jasonvarga:
I agree that we need to mention that you should have a server set up in the prerequisites - and we can recommend Valet - but I don't think we should be explaining how to set up a server.

pauldoerwald:
That would have been enough to help me. Like I said — the fact that I got a 403 in Chrome _implied_ that there was a server, and that's what sent me down the [fortunately shallow] rabbit hole.

PR #608, opened 2021-11-17
Update env variable to match

andymnewhouse:
The variable in the statamic/statamic repo is `STATAMIC_STACHE_WATCHER` not `STATAMIC_FILE_WATCHER`. Just think things should match to reduce confusion :)

https://github.com/statamic/statamic/blob/master/config/statamic/stache.php


jasonvarga:
Thanks!

PR #611, opened 2021-11-18
Update ubuntu.md

yorgos-dimitriadis:
Fix typo


jasonvarga:
Not really a typo, `vi` is a thing. But `vim` is probably better 👍

PR #612, opened 2021-11-19
Update nav.md

jcohlmeyer:
Make it more clear you need to use two colons when using the handle syntax. https://github.com/statamic/cms/issues/4708


jackmcdade:
Thanks!

PR #622, opened 2021-12-04
Add permalink variable to assets fieldtype docs

o1y:
I find the `permalink` variable very helpful when using the assets fieldtype. Couldn't find any info about the variable in the docs.


jackmcdade:
Thank you!

PR #623, opened 2021-12-05
Fix typo in redirect

Pantastisch:
The permanent redirect seems to have a typo.

Older references to [`statamic.dev/git-integration`](https://statamic.dev/git-integration) from config templates like `git.php` do lead to Mr Clippy 📎, which looks like incorrect. 😄


jackmcdade:
Nice catch, thanks!

PR #626, opened 2021-12-09
Updated Collection Tag Filtering Taxonomies

### LINKED ISSUE #625: Collection Tag Filtering Taxonomies does not refer to using variables for filtering
The example shown in [Collection Tag Filtering Taxonomies](https://statamic.dev/tags/collection#taxonomies) only refers to filtering by a literal string and I was not able to find an example of doing the same using a variable.

timgreenwood:
Added a description of how to use variables when filtering a collection by taxonomy

Fixes #625


jackmcdade:
That's not a feature of of the Taxonomy tag specifically, but all tags. It's documented in the Antlers section: https://statamic.dev/antlers#variables-inside-tag-parameters

PR #630, opened 2021-12-10
Add Swedish to available languages table

adevade:
I marked it as Work In Progress, as most of it is still machine translated. Feel free to remove if you want.

I also updated the count of languages at the top of the page.


jasonvarga:
Thanks!

PR #631, opened 2021-12-12
Add example for custom counter

zunnurs01:
Show the implementation with from and by parameters based on https://v2.statamic.com/tags/increment.


jackmcdade:
Good one, thank you!

PR #634, opened 2021-12-13
Update 'modify_date' content

zunnurs01:
- Update yaml syntax for 'now' to use v1.2 syntax
- remove "{{noparse}} {{/noparse}}"
- remove "{{ now | modify_date:....... }}" as is not working and replace with output


jackmcdade:
Thanks for bringing this up! We're no longer parsing the content of entries in the docs, and they're statically cached until the next deploy, so yeah that had to be fixed. I did it a bit differently though, and hit a few others while i was at it.

PR #635, opened 2021-12-13
Update forms.md

sebabal:
As far I can see "fieldtype" is actually just "type"


jackmcdade:
You are correct! Thanks!

PR #636, opened 2021-12-14
Move <ul> out of the taxonomy loop

Konafets:


jackmcdade:
Nice catch, thank you!

PR #638, opened 2021-12-15
Refer to the correct Markdown Library

Konafets:
In PR https://github.com/statamic/cms/pull/1181 Parsedown was substituted by League/CommonMark.


jackmcdade:
Good catch, thank you!

PR #641, opened 2021-12-18
Update git-automation.md

sjclark:
Add Ploi to providers which setup git remote and upstream branch automagically.


jackmcdade:
Thanks!

PR #645, opened 2021-12-18
Fix the reference link to Artisan url.

Konafets:


jackmcdade:
And thanks for this one too!

PR #653, opened 2021-12-19
Update converting-from-single-to-multi-site.md

sjclark:
Fix a copy-pasted typo (from above Structure section).


jackmcdade:
...and thank you!

PR #659, opened 2021-12-22
Update docs for multi-site configuration of cache

francoismajor:
Needs to be verified, but these changes have solved errors on my end and are what where proposed by Jason to me. So organizing by top level domains and using absolute urls for sites.

The hosts part I realized did not match with my dev environment, so I think this helps make it clearer. The path domain should match the server environment.


jackmcdade:
Thanks for this! Better late than never! 😅

PR #669, opened 2021-12-31
Update taxonomy.md

websmyth:
Make it clear that related entry data won’t appear in this example if the taxonomy isn’t linked to the collection. Appreciate the terminology in my blockquote might not be quite right...


jackmcdade:
Good idea, thank you!

websmyth:
Woop! Finally a contribution to the docs 😆

PR #670, opened 2021-12-31
Sort the variables alphabetically

Konafets:
For a better overview, sort the variables in alphabetical order in https://statamic.dev/variables.

### Before

---


jackmcdade:
Good one, thank you!

PR #687, opened 2022-01-14
Update `Content Queries` with new where methods

ryanmitchell:
Add:

orWhere()
whereNull()
whereColumn()

and 

Complex Where Clauses


jackmcdade:
Thank you!

PR #690, opened 2022-01-18
Fix line breaks in optimizing-assets

simonolog:
Fixes some unnecessary line breaks in https://statamic.dev/tips/optimizing-assets


jackmcdade:
Thanks!

PR #693, opened 2022-01-19
Add missing punctuation.

Konafets:


jackmcdade:
Thank you! Every dot counts.

PR #694, opened 2022-01-20
Explain that blueprints can be hidden

klickreflex:
Explain that additional collection blueprints can be hidden from the new entry menu.


jackmcdade:
Thanks!

PR #697, opened 2022-01-22
Correct blueprint validate key

Konafets:
The field key for adding validation rules is `validate`. In some place `validation` was used. Additional the wrong type was documented. Corrected it to array.


jackmcdade:
Thank you!

PR #700, opened 2022-01-24
Referring to the correct example in the screenshot.

Konafets:


jackmcdade:
Good catch, thank you!

PR #701, opened 2022-01-24
Bring some color the the options for UserRole and UserGroup field types

Konafets:
Use backticks instead of single quotes to be consistent with User field type.


jackmcdade:
Thanks!

PR #709, opened 2022-01-26
Update options for Markdown Fieldtype

Konafets:
- Add antlers
- Add parser
- Rename `restrict_assets` to `restrict`


jackmcdade:
Good one, thanks!

PR #720, opened 2022-01-28
`ray` modifier missing title

adevade:
Removed subtitles **Overview** and **Example**, since it's not present on most other modifier pages.

Also added related entries: 
- `Dump`
- `Console Log`


jackmcdade:
Thanks! Just FYI code blocks are `antlers` by default, so it's not really necessary to add it :)

PR #731, opened 2022-02-09
Modifiers | The backslash in the addSlashes example was not displayed.

Konafets:
For better clarity, I removed the parenthesis in favour for inline code.

Before:

After:


jackmcdade:
Works for me, thanks!

PR #733, opened 2022-02-09
Fix the title of the compact modifier.

Konafets:
The compact modifier has the title "As" which leads to two entries of "As" in the left navigation:

Fix also some copy.


jackmcdade:
Thanks!

PR #754, opened 2022-03-11
Updated TipTap website url

arlindmusliu:
I was learning more about Bard, TipTap and ProseMirror when I noticed that the link for TipTap doesn't work.


jackmcdade:
Good catch, thanks!

PR #761, opened 2022-03-18
Bumping the PHP requirements

vannut:
As per [the release blog](https://statamic.com/blog/statamic-3.3) the minimum php version is 7.4

Although laravel 8 is supported, which [supports php 7.3 or php 8](https://github.com/laravel/framework/blob/8.x/composer.json#L18), I changed it to 7.4 in the docs.


jackmcdade:
Thank you!

PR #762, opened 2022-03-19
Add Example with variable

zunnurs01:


jackmcdade:
Thanks! Made a few changes to follow recommend code syntax and removed the example that only works in the new parser – for now.

PR #771, opened 2022-03-28
Explain variable scoping

robdekort:


jasonvarga:
I'm not quite sure what this is explaining. What trouble were you running into?

If you do `{{ thisdoesntexist }}` it'll just output nothing.

robdekort:
That if you define variables in a partial you can't use them somewhere else. They have to be defined before calling in the partial. Same goes for variables in tag pairs. I didn't know and it had me very confused. The tag pairs I understood but the partials not. It makes sense. I just didn't know :-)

robdekort:
I thought it was very important. But please explain it better if you agree!

PR #773, opened 2022-03-29
Partial tag and new Antlers Parser

espenlg:
Specify how to use the partial tag with the new experimental Antlers Parser.


jackmcdade:
This isn't true though, are you running into an issue with the `{{ partial:something }}` format?

espenlg:
The is a very good point. It is just in combination with variables as part of the partial name. Reject the PR and I'll try to adjust.

jackmcdade:
Ah yes, if you want to use variables you'll need to pass them through a parameter for interpolation. 🤘

PR #780, opened 2022-04-06
Mention additional Static Caching settings in Multi-Site docs

stoffelio:
On multiple occasions people on Discord (as well as I myself at one point) got confused about using Static Caching on a Multi-Site. The correct settings are only explained in the docs for Static Caching, not in the ones for Multi-Site, which is why some people miss that vital information when creating a Multi-Site.

So I added a quick link to the bottom of the Multi-Site docs to avoid this confusion in the future.


jackmcdade:
Thanks!

PR #792, opened 2022-04-14
Accessing other field values from custom fieldtype

jesseleite:
I see a few problems with using these `values` though...

1. Hidden fields (using conditions) will still show up in this `values` payload, which may lead to unexpected consequences.

2. Bard values are JSON encoded, which adds unnecessary complexity for people wanting to parse through bard values.

Maybe we could `provide()` a normalized values payload (using some of the logic from our [Omitter](https://github.com/statamic/cms/blob/3.3/resources/js/components/field-conditions/Omitter.js) class to solve the above problems), and allow the user to inject those normalized values into their components via `inject: ['formValues']`?

Or maybe we should just apply these normalizations to the `values` in the store?

Thoughts @jasonvarga?

### DIFF
--- a/content/collections/extending-docs/fieldtypes.md
+++ b/content/collections/extending-docs/fieldtypes.md
@@ -296,6 +296,20 @@ public function augment($value)
 
 [Read more about augmentation](/extending/augmentation)
 
+## Accessing Other Fields
+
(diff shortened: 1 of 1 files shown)


jasonvarga:
These are just the docs. It explains how to do what's available now. It's perfect. 👌 

If someone runs into one of those problems, if it even is a problem, we can address it then in cms and update the docs.

PR #793, opened 2022-04-16
add valet for windows port as an option

damjanuzelac:
The docs don't mention there's a Valet port for Windows. This change fixes that.


jackmcdade:
Thank you!

PR #796, opened 2022-04-19
Fix grammar in antlers.md

oskarradon:
Line 405: "that they don't access to" -> "that they don't have access to"


jackmcdade:
Thanks!

PR #807, opened 2022-04-27
Conditional fields and Revealer fieldtype clarifications

jesseleite:
- [x] Clarify differences between revealer fieldtype and conditional field data flow.
- [x] Add documentation for `sometimes`, `required_if`, etc. conditional field validation helper rules.
- [x] Uncomment 'best practice' tip @jasonvarga left, now that we properly support this stuff since 3.3 👍

### DIFF
--- a/content/collections/docs/conditional-fields.md
+++ b/content/collections/docs/conditional-fields.md
@@ -264,3 +264,29 @@ Statamic.$conditions.add('...', ({ root, store, storeName }) => {
     //
 });
 ```
+
+## Validation
+
+If you wish to conditionally apply validation to conditionally shown fields, we recommend using the `sometimes` [Laravel validation rule](https://laravel.com/docs/validation#validating-when-present).
+
+```
+-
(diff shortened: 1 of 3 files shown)


content/collections/docs/conditional-fields.md
```diff
@@ -16,6 +16,14 @@ Field conditions are set on individual field settings in [blueprints](/blueprint
 
 You may specify various rules for showing a field under either the `if` / `show_when` keys, or hiding a field under the `unless` / `hide_when` keys.
 
+### Data Flow
+
+Only visible fields are submitted with your form data by default. This allows you to control data flow, and [conditionally apply validation](#validation) to visible fields when needed.
```
jackmcdade, line 21: It says `by default` but for now it's the only way.

jesseleite: True. That was in prep for the upcoming config, but I'll remove that for now.

Issue #812, opened 2022-05-05
`relate:` tag, not available anymore?


j3ll3yfi5h:
Just came across the relate tag. I think it is not available anymore?

https://github.com/statamic/docs/blame/master/content/collections/fieldtypes/terms.md#L86

joshuablum:
Hey @j3ll3yfi5h,

Thanks for point that out!
It's still available ([see code here](https://github.com/statamic/cms/blob/3.3/src/Tags/Relate.php)) but the docs for it are missing. Will fix that 🚀

PR #819, opened 2022-05-13
Add relate tag

joshuablum:
Mentioned in https://github.com/statamic/docs/issues/812 regarding https://github.com/statamic/cms/blob/3.3/src/Tags/Relate.php.

Eager for feedback. Not entirely sure about features, more potential use-cases and phrasing.

### DIFF
--- a/content/collections/tags/relate.md
+++ b/content/collections/tags/relate.md
@@ -1,0 +1,32 @@
+---
+title: Relate
+description: Converts fields to terms.
+intro: The relate tag is used for converting fields to terms.
+stage: 
+id: 
+---
+## Overview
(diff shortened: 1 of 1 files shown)


jackmcdade:
I don't think we need/want to document this field as it's only there for backwards compatibility. Augmentation made everything this tag does obsolete.

dadaxr:
Sorry @jackmcdade  for the revival but as mentioned in the original issue (#812 ) the actual documention has already a broken link to that "relate tag" stuff : 

the broken link is : https://statamic.dev/tags/relate

it is mentioned in this paragraph : https://statamic.dev/fieldtypes/terms#without-taxonomizing

I think we should either document it, or rephrase the actual doc ! (For what it worth this whole paragraph seems unclear to me)

Issue #820, opened 2022-05-16
`is_array` modifier is missing


wm-simon:
I've tested `is_array` and it's working, but I can not find the modifier in the docs.

joshuablum:
It seems a little confusing, so I guess you're both right, sorry.
Probably better to document it 💯

But like you said, @duncanmcclean, [it's](https://github.com/statamic/cms/blob/3.3/src/Modifiers/CoreModifiers.php#L1007) just a wrapper around PHP's function.
```php
public function isArray($value)
{
    return is_array($value);
}
```

joshuablum:
https://github.com/statamic/docs/pull/826

PR #826, opened 2022-05-18
Add `is_array` modifier

joshuablum:
Part of the core modifiers but not documented yet.
Mentioned in https://github.com/statamic/docs/issues/820.


jackmcdade:
Thanks!

PR #835, opened 2022-05-27
Adding next and prev to the nav

austriker27:
Adding details about `next` and `prev` to the Navigation page.

I learned about `next` and `prev` from @JohnathonKoster in the [Statamic Discord](https://discord.com/channels/489818810157891584/489819906540568593/979845502680055908). I think it's a useful thing in the docs - to save futureDavid (or anyone else) time.


jasonvarga:
Thank you, but this should only go on the `/new-antlers-parser` page somewhere for now.

We're not documenting new parser stuff outside of that page until it's the default.

austriker27:
> Thank you, but this should only go on the `/new-antlers-parser` page somewhere for now.
> 
> We're not documenting new parser stuff outside of that page until it's the default.

Oops, I didn't realize this was specific to the new parser. Thanks! I'll try to find a place for it.

PR #849, opened 2022-06-14
Passing Context into Blade Components

### LINKED ISSUE #847: Passing Context to `<x-layout />`
Howdy!

I was troubleshooting a problem with the Statamic SEO Pro plugin last week, and I came across this [fantastic little nugget](https://github.com/statamic/seo-pro/issues/172#issuecomment-825926742) from @thomasjohnkane (summarized below):

```blade
{{-- from any Statamic template page --}}
<x-layout :context="$__data">
</x-layout>
```

Setting this value lets you use `@aware` for any child blade component, significantly cleaning up those blade views.

I'm not sure if this is a pattern that the Statamic team wants to support, but if so, it might be worth featuring it in the page about Blade templates. What do you think?

theutz:
Closes #847

### DIFF
--- a/content/collections/docs/blade.md
+++ b/content/collections/docs/blade.md
@@ -251,6 +251,44 @@ First, add a `context` prop to your layout component.
 </html>
 ```
 
+Then, merge the `context` prop with the parent data in your Layout component's `data` method.
+
+```php
+<?php
+
+namespace App\View\Components;
+
+use Illuminate\View\Component;
+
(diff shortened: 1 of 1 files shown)


jackmcdade:
@jasonvarga, this topic of handling proper `context` inside of Blade has come up a number of times. Would you consider this a good solution — at least for now?

jasonvarga:
Just tried it. Seems like a good solution to me!

PR #856, opened 2022-06-27
Where to set the static cache's expiry time

joshuablum:
Point to the `expiry` key in the `config/statamic/static_caching.php` config file.


jackmcdade:
Good one 👍

PR #857, opened 2022-06-28
Document container option for link fieldtype

malteriechmann:
This change will add a documentation for the existing option `container` inside of the `link` fieldtype. It will also remove a not needed heading »Config options« from the documentation.


jackmcdade:
Thank you!

PR #859, opened 2022-07-02
Update blade.md

amhrn:
`$tag` is an array and gives an error **Attempt to read property "pages" on array** with `$tag->pages`


jackmcdade:
Thanks!

PR #864, opened 2022-07-05
Add section about dynamic CSRF tokens when using static caching

joshuablum:
CleanShot 2022-07-05 at 5 15 46


jackmcdade:
Awesome, thanks!

PR #877, opened 2022-07-21
Make the findBySlug example clearer

MohammadAbusaleh:
Make the findBySlug example clearer as the current example is a bit ambigious, which is slug and which is collection


jasonvarga:
Hey thanks for the PR.

`findBySlug` is deprecated so I'd actually prefer if it wasn't in the docs.

I've just pushed up an edit with a code example that shows how you can get an entry by slug.

64b97d681efc6a3fafaaf6a52d598d8545ad2eaa

PR #878, opened 2022-07-22
Theme (dark mode) toggle

jasonvarga:
Gives the option to match the system preference, or explicitly use light or dark themes.

### DIFF
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -43,5 +43,23 @@ window.bodyData = function() {
     };
 }
 
+window.htmlData = function() {
+    return {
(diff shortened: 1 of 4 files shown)


Konafets:
Hey, 

thanks for adding the toggle. Just tried it with Chrome and the toggle switches to light mode, but the setting is lost after reloading the page.

jasonvarga:
I renamed some variables at the last second and it broke that without noticing. Derp. I've fixed this in 35f6853afbf7be3993d5800895825bca3f44b2e6, sorry about that.

PR #892, opened 2022-08-06
fix spelling of 'controlling'

bmartus:
A quick spelling fix: 
  'version controling' => 'version controlling'


jackmcdade:
Thanks!

PR #899, opened 2022-08-20
Add information on forcing HTTPS

katjapvl:
It feels like the this shouldn't be necessary, but for whatever reason, assets generated with Vite on DO App Platform insist on using `http://`. This persists through setting `APP_URL`, `APP_ENV`, etc.

Or maybe there's a better way, and I've just missed it.

### DIFF
--- a/content/collections/docs/digitalocean-app-platform.md
+++ b/content/collections/docs/digitalocean-app-platform.md
@@ -106,6 +106,21 @@ Navigate to the Settings tab of your application and add your commands to the Bu
     <img src="/img/deploying/digital-ocean-app-platform/build-command.png" alt="Authorizing Github">
 </figure>
 
+## Forcing HTTPS
+
+When using the [vite tag](https://statamic.dev/tags/vite) instead of Mix, built assets will be served from `http://` by default. To address this, HTTPS can be forced from `app/Providers/AppServiceProvider.php`:
+
+```php
+public function boot()
+{
+    ...
+
+    if (env('APP_ENV') !== 'local') {
+        URL::forceScheme('https');
+    }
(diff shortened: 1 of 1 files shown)


jackmcdade:
This definitely feels wrong to me. Let’s pursue a different solution! 😅

katjapvl:
Agreed.

Similar issues have been ongoing for some time. Two quick examples:

https://statamic.com/forum/5080-hosting-on-digitalocean-app-platform-mixed-content
https://medium.com/smallworldus/deploying-statamic-3-to-the-digitalocean-app-platform-75e9252ec8e1

There are also non-Statamic issues reporting similar with Vite asset generation:

https://stackoverflow.com/questions/73036658/laravel-vite-sail-over-https-possible
https://stackoverflow.com/questions/68499958/laravel-shows-urls-with-http-instead-of-https
https://stackoverflow.com/questions/35827062/how-to-force-laravel-project-to-use-https-for-all-routes
https://shouts.dev/articles/force-laravel-to-use-https-connection

Strangely, the Statamic Vite gives examples using `http` https://statamic.dev/tags/vite. I don't see any option to do something like `npm run build -- --https`. 

I'll have to mess around more.

katjapvl:
@jackmcdade I'm not having issues with this anywhere _except_ Digital Ocean App Platform, so this seems to be specific to whatever it is they're doing / some oddity with Laravel.

I'm not sure how worth investigating the 'why' is. I'm just going to stick to Forge + Droplet.

jackmcdade:
Weird. Maybe a support ticket to DO could help? Dunno! Gonna close this for now then :)

PR #920, opened 2022-09-22
Update collection-repository.md

JuanS:
There is a typo, the doc refers to `$container` while it should be `$collection`


joshuablum:
Gracias!

PR #925, opened 2022-10-05
Changed structure to nav permission in users.md

dagy304:
The permission for structures got changed to navs but was not updated in the docs.


jackmcdade:
Thanks!

PR #933, opened 2022-10-12
change code block type to shell

manogi:
I know it is a very small betterification, but in nagged me for a while now ;-)


jackmcdade:
Thanks!

PR #934, opened 2022-10-15
Added php please assets:generate-presets to docs

ChristopherDosin:
If you have changed the presets config, but you already uploaded the images you want to re-generate the images.


jackmcdade:
Good one, thank you!

PR #938, opened 2022-10-17
Explain setting up a git remote for git automation

jesseleite:
This is already done for Forge users, Ploi users, etc, but maybe we could better explain how to setup a git remote for other deployment strategies that don't do this.


jackmcdade:
Nice 👌

PR #940, opened 2022-10-25
fix wrong link

simonvomeyser:
I added what I believe to be a link making more sense, since `/tips/storing-entries-in-a-database` explicitly talks about how to manage Eloquent models within Statamic

If I'm mistaken here feel free to simply close this PR, just was totally confused as I followed the link.

Best
Simon


jackmcdade:
Yes, this is much better. That link was all we had to refer to back in the day before any of the new driver stuff. Thanks!

PR #941, opened 2022-10-28
Fix wrong path

helloDanuk:
I think the path to the language folder changed with a newer laravel version.


jasonvarga:
It's not wrong, just old. Both paths still work.

PR #944, opened 2022-10-31
Correct relative modifier docs, include example output

jacksleight:
The relative modifiers `remove_modifiers` param should be `true` to remove the extra words, as per [the comment](https://github.com/statamic/cms/blob/5082efc80f2bf22416c79649542ee501be7a9e34/src/Modifiers/CoreModifiers.php#L1942).

However the docs actually say that you should use `false`.

This PR corrects that and also adds some actual output examples.

Related to https://github.com/statamic/cms/pull/6976


jackmcdade:
Thanks!

PR #948, opened 2022-11-11
Add max_items to terms and collections docs

jacksleight:
The Terms and Collections fieldtypes support `max_items` but it's missing from the docs. This PR adds them.

Also the Collections fieldtype says it has a `collection` option, which it doesn't, so that has been removed.


joshuablum:
Thanks, Jack!

PR #956, opened 2022-11-29
Add `length` as a reserved word

Konafets:
As mentioned in https://github.com/statamic/cms/blob/972255aa973463cb3543fc3a10701bfe90e34541/src/Http/Controllers/CP/Fields/FieldsController.php#L78 the word `length` is also a reserved word.


Konafets:
@jasonvarga Fastest merger ever.

PR #957, opened 2022-11-30
Update forms.md

Sennik:
Adding cache related information as form validation information is not returned if caching is used on the page where the form is used.


jackmcdade:
Thanks!

PR #964, opened 2022-12-09
Include .env database configuration in instruction

drese:
The default/fresh Statamic installation does not include DB environment variables. Added a step in the instructions to include this, prior to migration.


jasonvarga:
They're definitely in there:

https://github.com/statamic/statamic/blob/da2f408fb219534b707481b04c33c4e6acea0fe0/.env.example#L19-L24

PR #965, opened 2022-12-09
Update static-caching.md

stuartcusackie:
Add notes about static:warm.
I just read about this command elsewhere. It's not mentioned in the docs. Feel free to change my text.
I've also heard a mention of being able to define the queue (warm_queue). Notes for this should be added too.


jackmcdade:
Thank you!

PR #972, opened 2023-01-04
Correct fully qualified, absolute URL code example

stefankempf:
The currently shown version `env('APP_URL').'fr/'` is not working for me, the corrected one is. Not sure if this is the right approach / best practise but proposing the change anyway.


jackmcdade:
The default url is `/`, so if we change the example to `/fr/` folks following along will end up with a URL of `//fr/` which is a bit broken. You can feel free to set it however you want, with the trailing slash in either place. I think it's still clear enough for folks to figure it out. Thanks for the suggestion though!

PR #976, opened 2023-01-10
Fix typeo in new-antlers-parser

marzvrover:
Just a quick typeo fix.

The example of rendered output had an extra <kbd>"</kbd>


jackmcdade:
Thanks!

PR #977, opened 2023-01-10
Highlight, that Live Preview does not work with `array` cache driver

jakubjo:
I'll keep the description short, since there is nothing more to say. :-)


jackmcdade:
Thanks!

PR #983, opened 2023-01-16
Update cp-translations.md

MarvelousMartin:
Czech translation is available since https://github.com/statamic/cms/pull/6320


joshuablum:
Thanks!

PR #984, opened 2023-01-16
Fix broken link in Modifier ->Deslugify

MrSubidubi:
The link in "deslugify" to "dashify" does not work. Removing the "#" should fix the issue,


jackmcdade:
Thanks!

PR #988, opened 2023-01-25
Update the description about valid variable names

Konafets:
The copy says that hyphens are allowed within a variable name but the regex does not reflected that and after a chat with @JohnathonKoster it turned out that hyphens at the end of a variable name are not allowed. I rephrased the paragraph and added the complete regex which describes a variable name.


jackmcdade:
Because of the conflict i had to push this on a new commit. Thanks for the PR!

PR #989, opened 2023-01-26
Add bool_string modifier

el-schneider:
Mostly for my own sanity, as I tried to use `to_bool` without success over and over 😃


jackmcdade:
Great catch, thank you!

PR #990, opened 2023-02-01
Update docker.md

joseph-d:
Added a tip for people to run `sail artisan statamic` instead of `php please` when using sail as it really isn't obvious


jackmcdade:
Nice one, thanks!

PR #997, opened 2023-02-13
Add bard inline config, clarify whats supported

jacksleight:
Docs for https://github.com/statamic/cms/pull/6973

See also https://github.com/statamic/cms/issues/7530


joshuablum:
Thanks!

PR #1004, opened 2023-02-26
Fix wrong path in readme

bambamboole:
I just followed this install guide. The `users` folder was created in `resources`


joshuablum:
Good catch, thanks!

(If using Statamic with flat-file users the `users` directory at the root contains the`.yaml` files for them. `resources/users` is the place for the roles and groups config 👍)

PR #1008, opened 2023-03-08
Added missing '=>' to function declaration

joshuaanderton:
Minor fix; added missing '=>' to function declaration in hooks example.

P.S. Huge fan of Statamic!


jackmcdade:
Thanks @joshuaanderton !

PR #1011, opened 2023-03-15
Update quick-start-guide.md

gmgale:


joshuablum:
Good catch, thanks @gmgale!

PR #1012, opened 2023-03-16
Add example for $term->fromDataLocale(), using default settings

jarnalyrkar:
Have a code example for the default single site settings, and provide an example of what data to pass into the second argument.


joshuablum:
Thanks for the PR, @jarnalyrkar! Did some minor tweaks and adjustments.

PR #1013, opened 2023-03-16
Update iso_format modifier doc

robdekort:
Add localization options using macro-formats.


robdekort:
Ignore what-the-diff cuz what-the-f?

joshuablum:
How neat is that? Thanks, @robdekort!

robdekort:
Happy to. If only I knew before ;-)

PR #1015, opened 2023-03-21
add 'save' to the list of reserved words

mnlmaier:
After hours of almost crying, I was able to pinpoint the following issue to the accidental use of the handle `save` in an entry.

---

When used, entries will trigger Events when being opened in the CP and also when accessed in the frontend. This pollutes the queue and might cause some headache.


joshuablum:
Thanks @mnlmaier, good catch! Will prevent others from crying then in the future 👍

PR #1019, opened 2023-03-30
Update sanitize.md

ewhicher:


joshuablum:
Hey @ewhicher,

Thanks for taking the time and the PR!

The way it is described in the docs right now is actually correct since they refer to the Antlers `runtime` parser that is the default since Statamic 3.4 (and was available as an option since 3.3). The docs for the "old" `regex` parser are describing it the way you're proposing.
See the difference [here (legacy `regex`)](https://statamic.dev/antlers-legacy#modifying-data) and [here (new `runtime`)](https://statamic.dev/antlers#modifiers).

Therefore I'm going to close this PR since it refers to and outdated syntax.

ewhicher:
Ahh, my parser was set to 'regex' as it did not update when I upgraded. As it was said to be the 'default' I just assumed that running the update would do this for me.

Thanks!

PR #1037, opened 2023-05-02
Update collections.md

benfurfie:
Add in an explainer of how to loop through mounted collections dynamically without needing to define the collection (useful for partials used across multiple collections/pages).


jackmcdade:
Nice one, thank you!

PR #1048, opened 2023-05-17
Fix incorrect CSS path in code example

hybridvision:
The path for cp.css was incorrect in the example code 🙃


joshuablum:
Thanks, @hybridvision!

PR #1055, opened 2023-05-29
Edit utilities.md

MarvelousMartin:
I have upgraded to Statamic 4.0 from 3.4 and Utility::make('')->...->register() did not work, so with help of @edalzell changed it to
```php
$utitlity = Utility::make('')->...;
Utility::register($utility);
```
It works like a charm now.

### DIFF
--- a/content/collections/extending-docs/utilities.md
+++ b/content/collections/extending-docs/utilities.md
@@ -21,8 +21,20 @@ Registering a utility will give you a route, nav item, and a permission for free
 
 In a service provider's `boot` method, you can register a utility with the `Utility` facade.
 
-Start with `Utility::make()`, chain as many methods as you want, then finish with `register()`. At a minimum, you'll need a view.
+Start with `Utility::make()`, chain as many methods as you want, then finish with `Utility::register($utility)`. At a minimum, you'll need a view.
+```php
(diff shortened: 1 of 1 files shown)


MarvelousMartin:
I don't know now with the statement `At a minumim, you'll need a view`. I don't have a view() there and yet it still works. Feel free to update it furthermore.

jasonvarga:
The correct way is explained in the code block directly after.

CleanShot 2023-05-29 at 15 11 28

I'll update the rest of the text. It looks like we missed a couple of bits.

MarvelousMartin:
Just one more thing:
I get `Method 'extend' not found in \Statamic\Facades\Utility` in my code. So I will have it the previous way. Or I am doing something wrong, I don't know..

jasonvarga:
Is that an actual error? Or is it just your IDE hinting that the method isn't there?

MarvelousMartin:
Hinting, but even though I don't see the extend method, it works. Sorry for the confusion.

edalzell:
Ya that's a docblock thing on the Facade.

jasonvarga:
Making these docs changes in #1056, and making your facade typehint change in statamic/cms#8213.

PR #1058, opened 2023-05-31
Add explanation how to access config values

manogi:


jackmcdade:
Nice extra detail, thanks!

PR #1069, opened 2023-06-19
Make it more obvious that local install creates a new subdirectory

wturrell:
(because I think the wording is a bit ambiguous)


jackmcdade:
Fair enough, thanks!

PR #1070, opened 2023-06-21
Update example URL for asset tag

wturrell:
*Might* be less confusing for new users - path matches the upload directory when using the default asset container in a new Statamic install.


jackmcdade:
We're not using the default location on purpose to show you that you can pass _any_ URL into it. :)

PR #1073, opened 2023-06-27
is external url modifier docs

martyf:
Just a simple docs page for the new `is_external_url` modifier [statamic/cms PR 8351](https://github.com/statamic/cms/pull/8351)

Follows the same setup as the `is_url` entry but also not sure if the code samples should be improved. 

The antlers, as is, is invalid and doesn't return the HTML the docs suggest.

Could/should this (and `is_url`) be re-worked to be more copy-and-paste complete? Or is this adequate for the use case. It makes total sense... but just questioning whether it needs greater detail. Or even a use case.

Happy to discuss and update as needed.

### DIFF
--- a/content/collections/modifiers/is_external_url.md
+++ b/content/collections/modifiers/is_external_url.md
@@ -11,14 +11,17 @@ Returns `true` if a string is an external URL.
 ```yaml
 google_url: http://google.com/
 entry_url: /waffles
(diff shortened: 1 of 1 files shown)


jasonvarga:
IMO the code example is not intended to be copy-pasteable here. It's just enough to show the true/false outcomes.

Thanks for this!

PR #1077, opened 2023-06-28
Fixed wrong default table name

jure-knezovic:
Hello there!
While trying to install Statamic 4 into existing project, i followed this guide. However, database name is different from one created by php please auth:migration. This change fixes this issue.


joshuablum:
This got changed in https://github.com/statamic/cms/pull/8059.

Thanks, @jure-knezovic!

PR #1079, opened 2023-07-03
Update vite.md for vite:asset

conradfuhrman:
An extra colon was present for the example code using vite:asset


joshuablum:
Good catch, thanks Conrad!

PR #1087, opened 2023-07-13
Clarify location of directory in icon fieldtype options

tenmillionteeth:


jasonvarga:
It's not in your `resources` folder. It's from the root of your project.

PR #1092, opened 2023-08-20
Update control-panel.md with externalScript reference

caseydwyer:
Add in a reference to the `externalScript` method available on `Statamic\Statamic`.


jackmcdade:
Great addition, thank you!

PR #1094, opened 2023-08-26
[4.x] Update cp-translations.md: add update instruction

jeroenpeters1986:
Added a little bit of explicit update instructions for those wondering

### DIFF
--- a/content/collections/docs/cp-translations.md
+++ b/content/collections/docs/cp-translations.md
@@ -89,6 +89,16 @@ resources/lang/
 |-- eo.json
 ```
 
+#### Updating the translation files
+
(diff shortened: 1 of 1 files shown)


jasonvarga:
Thank you for that, sorry for the delay. I've instead put a note on the page to explain the same command will update the files.

jeroenpeters1986:
That’s also great! If there is just a little pointer in that direction it’s helpful 😄Op 17 jan 2024 om 22:06 heeft Jason Varga ***@***.***> het volgende geschreven:﻿
Thank you for that, sorry for the delay. I've instead put a note on the page to explain the same command will update the files.

PR #1095, opened 2023-08-29
Reserved Words: Add `count` to reserved words list

AlexanderFalkenberg:
Using **count** as a field name in a field set (e.g. show limited number/count of entries), which might be used inside a loop e.g. repeater field can have unintended consequences.


duncanmcclean (approved):
Thanks!

PR #1096, opened 2023-08-29
Update nocache.md to reflect usage with default Laravel routes

andjsch:
Add an info for the usage with Laravel Routes


duncanmcclean:
Thanks for this PR!

However, we're planning to fix this in Core soon so this workaround won't be needed. See statamic/cms#9406.

PR #1098, opened 2023-09-04
Update quick-start-guide.md

sunscreem:
Now vite is the default `npm run watch` needs updated to `npm run dev` (I think!)


joshuablum:
We should probably overhaul the whole guide now but in the meantime this helpful.

Thanks!

PR #1102, opened 2023-09-13
add missing parameter to published() call

dadaxr:
In the doc example we are chaining calls to create an Entry : 
```
Entry::make()
            ->published()
            ->data(['title' => 'About us', 'subtitle' => 'We are awesome'])
            ->etc(); // and so on...
```

but the "published()" method return a boolean if no parameter is passed, so the "->data()" chained call breaks.

When passing a parameter to "published()", "this" is returned so the chained call works.


duncanmcclean:
Good catch, thanks!

PR #1103, opened 2023-09-16
[4.x] Adds tip for the default icon set

bretterer:
This PR adds a "tip" section when talking about Icons to direct people to the available icons that can be used.


jackmcdade:
Good idea, thank you!

PR #1105, opened 2023-09-25
Collections: Add more examples and detailed explanation for collection routing

dadaxr:
This PR tried to achieve 2 goals : 

- insist more on the fact that we can use, not only meta variable in collection routes but also, fields defined in the blueprint
- suggest that for advanced stuff we can also use field using computed values.

Why ? because I had this project with a "products" collection and a "categories" collection (each product belonging to a category). For SEO purpose, product url was supposed to be "/category_url/product_slug" or even "/master_category/child_category/product_slug" (when using nested categories) .

I had hard time figuring out how to achieve that while reading the doc carefully. 

I first attempted to recreate custom routes, but it was quite complicated and implied a lot of side effects (like product urls not included in seo-pro sitemap ...)

then I tried to use Antlers in my routes : 

``` yaml 
route: "{{ category.url() }}/{{ slug }}"
```

But it triggers an error, so my last attempt was to use that computed values, and it worked beautifully (without side effects).

### DIFF
--- a/content/collections/docs/collections.md
+++ b/content/collections/docs/collections.md
@@ -227,8 +227,7 @@ For non-structured collections, you can choose which field and direction to sort
 
 ## Routing
 
-Entries receive their URLs from their collection's route setting. You can use standard meta variables in addition to the variables from the collection's blueprint to define your route rule.
-You can also use [computed values](/computed-values) or even use Antlers syntax to achieve advanced stuff.
+Entries receive their URLs from their collection's route setting. You can use standard meta variables in addition to the variables from the collection's blueprint to define your route rule. You can even use [computed values](/computed-values) or Antlers to do advanced things.
 
 ``` yaml
 route: /blog/{slug}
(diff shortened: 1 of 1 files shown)


duncanmcclean (approved):
Thank you!

content/collections/docs/collections.md
```diff
@@ -281,29 +280,29 @@ route: /{parent_uri}/{slug}.html
...
-// define the computed value in a service provider
 Collection::computed('products', 'category_url', function ($entry, $value) {
     return $entry->category?->url();
 });
 ```
 
 ``` yaml
 route: '{{ category_url }}/{{ slug }}'
-# example: /wooden-toys/steam-locomotive
+# example: /categories/wooden-toys/steam-locomotive
```
dadaxr, line 305: @duncanmcclean , actually the /categories/ prefix here does not match the pattern given in the yaml route above.
either you add it to the route defintion : 

`route: 'categories/{{ category_url }}/{{ slug }}'`

or you remove it from your example.

what do you think ?

duncanmcclean: I added it to try and make it clearer that the `/categories/wooden-toys` part *was* the `{{ category_url }}` and you're just appending the slug to the end of it.

PR #1110, opened 2023-10-03
Update ensure_right.md

hybridvision:
Fix code example that was missing the pipe separator.


joshuablum:
Good catch, thanks @hybridvision!

PR #1113, opened 2023-10-10
Update antlers.md

SDenny:
Fixes typo in modulus assignment operator description.


joshuablum:
Good catch, thanks!

SDenny:
> Good catch, thanks!

Any time!

PR #1116, opened 2023-10-16
Update search.md

robdekort:
H3 instead of h2


joshuablum:
### Good catch, thanks!

PR #1145, opened 2023-11-29
Add lang option to config example on multi-site.md

DanielDarrenJones:
A lang option was added to enable keeping a general locale but using a difference set of language files, this isn't currently ver obvious from the docs, but as a simple first step adding it into the example config could help. 

We could potentially show this as commented out with a note about it being optional?

### DIFF
--- a/content/collections/docs/multi-site.md
+++ b/content/collections/docs/multi-site.md
@@ -30,7 +30,8 @@ return [
     'sites' => [
         'default' => [
             'name' => config('app.name'),
-            'locale' => 'en_US',
+            'locale' => 'en',
+            'lang' => 'en_US',
             'url' => '/',
             'direction' => 'ltr',
         ]
--- a/content/collections/docs/multi-site.md
+++ b/content/collections/docs/multi-site.md
@@ -31,9 +31,9 @@ return [
         'default' => [
             'name' => config('app.name'),
             'locale' => 'en_US',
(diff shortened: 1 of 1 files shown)


jasonvarga:
You've switched them.

`locale` should be the more specific `en_US` and `lang` should be just `en`.

DanielDarrenJones:
Based on https://github.com/statamic/cms/issues/4502#issuecomment-954844836 The Locale will always be shorted to the  `short_locale` so it would be the lang which is the full one wouldn't it?

I have updated the text here to be a bit more clear (hopefully).

jasonvarga:
The lang is usually the short one - it corresponds with the language file. `en.json`, etc.

By default, the lang will be the shortened version of locale.

Only in his very specific example did he need the long version which was the reason we added the feature. He needed `de_CH` which is one of the few regional language files. See https://github.com/statamic/cms/tree/4.x/resources/lang

DanielDarrenJones:
We are currently hitting this on a multisite install as well, as we have a UK and US version of the site (they are extremely similar in text, but with some subtle differences). So we needed this feature but struggled to find out about it, even with reading this section of the docs. 

Hopefully the simplified change I have done makes this more obvious now?

jasonvarga:
I've made additional changes. Let me know if it makes sense and I'll deploy it.

DanielDarrenJones:
Much better, thanks Jason!

PR #1147, opened 2023-12-01
Update forms.md

robdekort:
I don't think this line should go as in the examples the js_driver is being used, but @jasonvarga should be the judge of this.


jasonvarga:
That's intentional because you're replacing the default alpine driver x-model with a custom one.

PR #1162, opened 2023-12-09
Update command-not-found-statamic.md

esmiles711:
Updated composer location in the ~/.config file as of Laravel 5.1.


jasonvarga:
Thank you, but I don't believe that is correct. If it is, maybe you have a unique setup.

I'm happy to be wrong about that though. Please link to documentation explaining how you got composer installed into the .config directory.

PR #1172, opened 2023-12-18
Update GraphQL documentation to include explicit GraphiQL tip

highpanda:
Includes a tip that GraphiQL is available in the CP when GraphQL is enabled.

As a user, I didn't notice GraphiQL in the CP until after I had already built my queries manually. This would have been really helpful.


joshuablum:
Thanks @highpanda, neat tip!

PR #1205, opened 2023-12-19
Update fieldtypes.md with Fieldtype Category info

caseydwyer:
This PR adds context on adjusting Fieldtype Categories, which allows custom fieldtypes to be moved between the various categories, as defined in `components/fields/FieldtypeSelector.vue`.


jackmcdade:
Thank you!

PR #1206, opened 2023-12-26
Update tip regarding PHP versions on Netlify

ancaemcken:
Netlify uses PHP 8.0 by default. If `composer.json` requires another version, the build process for static site generation fails because of divergent PHP versions between the project and the Netlify environment. 

To solve the problem, set the PHP buildtime version on Netlify using the `PHP_VERSION` environment variable, as per [Netlify documentation](https://docs.netlify.com/configure-builds/available-software-at-build-time/).

### DIFF
--- a/content/collections/docs/netlify.md
+++ b/content/collections/docs/netlify.md
@@ -15,7 +15,7 @@ Deployments are triggered by committing changes to your Git repository. Alternat
 ## Prerequisites
 
 :::tip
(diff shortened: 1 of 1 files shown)


joshuablum:
Thanks, @ancaemcken! Netlify should _really_ upgrade their supported PHP versions 👀

intrepidws:
> Thanks, @ancaemcken! Netlify should _really_ upgrade their supported PHP versions 👀

Thanks to quite a bit of nagging, they finally have! I just sent over PR #1223 to update the docs.

PR #1222, opened 2024-01-14
Link added to the Laravel upgrade guide

vhbelvadi:
There’s a Laravel upgrade guide on Statamic.dev not listed on the upgrade guides page. It would be helpful to list it here.


joshuablum:
Hey @vhbelvadi,

Thanks for the PR! The guide covers Laravel 8 to 9 which we don’t officially support anymore since Statamic 4 (requires Laravel 9+). So it’s kinda outdated. I think it’s better to link to Laravel’s own upgrade docs where people can then look at the docs for their version.

Closing this in favor of mentioning Laravel’s own docs on upgrading.

PR #1228, opened 2024-01-18
Update optimizing-assets.md

SilverComet89:
Added in a link to the Nav > Performance section as it took me almost an entire full day before being aware of this section as I pulled my hair out trying to work out why it was taking 3 seconds for TTFB


jackmcdade:
Good one, thank you!

PR #1238, opened 2024-01-31
Recommend new `pro:enable` command instead

jesseleite:
For https://github.com/statamic/cms/pull/9435, if that gets accepted

Should we maybe also add a note for older installations where this new `pro:enable` command doesn't exist?


joshuablum:
🚢

PR #1248, opened 2024-02-11
Update data.md

benfurfie:
Fixes example. If the ID is 1, it would return 1, not 123.


duncanmcclean (approved):
Well spotted, thanks!

We use `123` as an example ID for other entries on that page so I've adjusted the entry's ID instead of the comment to stay consistent.

PR #1252, opened 2024-02-16
Add limit parameter to foreach tags

Spidlace:
The "**limit**" parameter is found in the "**foreach**" tags  but does not appear in the documentation.

=> https://github.com/statamic/cms/blob/4.x/src/Tags/Iterate.php#L47


jackmcdade:
Thank you!

PR #1262, opened 2024-02-27
fix customize commit commands

cwilsn:
Updates the customize commit command to keep the config name and email


cwilsn:
without the user/email in the git config our CMS edit commits were failing when in a deployed job

duncanmcclean (approved):
Thanks!

PR #1266, opened 2024-03-02
Fix Vercel Deploy Formatting

eiiot:
Changes formatting of code block to `bash`. At the moment on the current docs, it displays as "antlers."

CleanShot 2024-03-02 at 14 08 39@2x


jackmcdade:
Thank you!

PR #1302, opened 2024-03-31
Update docs on implementing custom Live Preview endpoint.

panda-madness:
The current implementation in the docs seems outdated, since no `LivePreview` facade seems to be available in Statamic. I've substituted that outdated bit with one that I've implemented and works on one of my sites.

### DIFF
--- a/content/collections/docs/live-preview.md
+++ b/content/collections/docs/live-preview.md
@@ -173,15 +173,19 @@ preview_targets:
 ```
 
 ```php
-use Facades\Statamic\CP\LivePreview;
+use Statamic\CP\LivePreview;
 use Illuminate\Http\Request;
 
-Route::get('/render-live-preview', function (Request $request) {
-  $entry = LivePreview::item($request->statamicToken());
-  $entry->title;        // The edited title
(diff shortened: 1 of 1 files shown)


duncanmcclean:
> The current implementation in the docs seems outdated, since no LivePreview facade seems to be available in Statamic.

While it's true that Statamic doesn't include a `LivePreview` facade, the example is actually using Laravel's [real-time facades feature](https://laravel.com/docs/master/facades#real-time-facades) so we can use the `Statamic\CP\LivePreview` class as if it _was_ a facade, without us actually needing to create one.

What issues were you running into when using the current example in the docs?

panda-madness:
> While it's true that Statamic doesn't include a LivePreview facade, the example is actually using Laravel's [real-time facades feature](https://laravel.com/docs/master/facades#real-time-facades) so we can use the Statamic\CP\LivePreview class as if it was a facade, without us actually needing to create one.

I've been programming with the Laravel framework for 6 years now and did not know that was a thing. If it works it works I guess. I'll close this out then, this PR is not relevant, sorry to bother you over nothing.

PR #1315, opened 2024-04-10
Minor bugfix to nav-breadcrumbs.md

sjclark:
Minor bugfix to example `figure` internal left margin

Screenshot 2024-04-10 at 8 40 43 pm


duncanmcclean:
Thanks!

sjclark:
> Thanks!

Just realised it should probably just be `space-x-4 mx-4` on the parent to remove all the existing inline margins, but same difference 🤷

duncanmcclean:
> Just realised it should probably just be `space-x-4 mx-4` on the parent to remove all the existing inline margins, but same difference 🤷

Done! :)

PR #1334, opened 2024-05-02
Static Caching: Update rewrite rules

duncanmcclean:
This pull request updates our documented static caching rewrite rules, to workaround various edge cases (as seen in statamic/cms#2619 and statamic/cms#9845).

The updated rules ensure:

* Only `GET` requests are served statically cached HTML files
* URLs with the `live-preview` query parameter will go straight to the `index.php` file, bypassing the static cache.

## Updates

* [x] Apache
* [x] Nginx (thanks @ryanmitchell!)
* [ ] IIS


duncanmcclean:
I'm going to merge this without updating the IIS config for now (since I don't have access to an IIS server to test on).  We can loop back to updating the IIS docs later.

PR #1336, opened 2024-05-13
Update local.md

benfurfie:
Add note on how to install Statamic 4 due to compatability issues with version 5 and some projects.

Need to make note to remove instructions probably around August once version 5 has had time to bed in and have all compatability bugs fixed.

### DIFF
--- a/content/collections/docs/local.md
+++ b/content/collections/docs/local.md
@@ -53,6 +53,10 @@ Next, you'll be prompted to set up your first super admin user. Do it.
 
 _After that_, everything is finished!
 
+:::tip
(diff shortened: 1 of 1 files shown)


duncanmcclean:
Thanks for this pull request!

Now that Statamic 5 is out, we're [no longer supporting Statamic 4](https://statamic.dev/release-schedule-support-policy#support-policy) (apart from for security fixes) so we don't want to be recommending developers start new sites with it.

> Add note on how to install Statamic 4 due to compatability issues with version 5 and some projects.

We've already fixed one compatibility issue where Statamic 5 couldn't be installed alongside PHPUnit 11. If you discover any other compatibility issues with Statamic Core and other packages, please open an issue and we can investigate.

PR #1340, opened 2024-05-17
Update docs on querying entry statuses

duncanmcclean:
This pull request adds a section to the "4 -> 5" upgrade guide around changes to querying entry statuses.

Related: https://github.com/statamic/cms/pull/9317


duncanmcclean:
@jasonvarga When you get a chance, would you be able to give this a look over? Feel free to tweak / re-word as needed.

PR #1341, opened 2024-05-19
Update collection.md

sunscreem:
I recently introduced a friend to Statamic and this section confused him. In the copy above it talks about a UL/LI example but the code example doesn't show them anymore. I think this is a better example for newcomers to follow with the UL/LI tags.

FYI - This was simplified in f0cc4bb7efb7f5bae457f7b8e2c38d97bbf0a9ab


jackmcdade:
Thanks!

PR #1342, opened 2024-05-20
Install on Ubuntu add php-curl

clementmas:
I followed the documentation: https://statamic.dev/installing/ubuntu
but I got an error during the `composer install` step because `php-curl` was missing. Adding it fixed the issue.


duncanmcclean (approved):
Thanks!

PR #1343, opened 2024-05-21
Update multi-site.md

ali-awwad:
Remove slash from APP_URL because it will create double "//" slashes in the site:url


duncanmcclean:
Thanks!

PR #1355, opened 2024-05-29
Update ubuntu.md

robdekort:
I bet 22.04 and 24.04 also work these days. Perhaps those need an update as well. No expert though.


duncanmcclean:
> I bet 22.04 and 24.04 also work these days. Perhaps those need an update as well. No expert though.

Looks like Forge supports both of them so I've updated the page. Thanks for this PR!

robdekort:
Nice that’s what I thought as well. Welcome!

PR #1359, opened 2024-05-31
Add example of how to access error by variable

Neophen:
I was looking for this information for a while until i figured it out by just trying different things. I though that it wouldn't hurt to have this information in the docs

### DIFF
--- a/content/collections/tags/form-create.md
+++ b/content/collections/tags/form-create.md
@@ -55,7 +55,7 @@ variables:
     name: error
     type: array
     description: |
-      An array of validation errors indexed by **field name**. For example: `{{ error:email }}`
+      An array of validation errors indexed by **field name**. For example: `{{ error:email }}`, to access the error with a variable use it like an object `{{ error[variable_name] }}`
   -
     name: old
     type: array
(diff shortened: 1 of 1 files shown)


content/collections/tags/form-create.md
```diff
@@ -55,7 +55,7 @@ variables:
     name: error
     type: array
     description: |
-      An array of validation errors indexed by **field name**. For example: `{{ error:email }}`
+      An array of validation errors indexed by **field name**. For example: `{{ error:email }}`, to access the error with a variable use it like an object `{{ error[variable_name] }}`
```
marcorieser, line 58: Not sure if that hint is needed since it's just how dynamic access for arrays in Antlers works (https://statamic.dev/antlers#dynamic-access)

duncanmcclean:
Thanks for this pull request!

However, like @marcorieser mentioned, this is just a "normal" feature of Antlers, nothing special to errors, so I don't think we necessarily need to mention it here.

content/collections/tags/form-create.md
Neophen: makes sense, just saying from a newcomers perspective.

PR #1362, opened 2024-06-09
Sort reserved words list alphabetically

bradfloodx:
Using `hook` as a field name throws `Too few arguments to function Statamic\Entries\Entry::hook()` errors when fetching collections via the API.

Also I've sorted the list alphabetically.

### DIFF
--- a/content/collections/tips/reserved-words.md
+++ b/content/collections/tips/reserved-words.md
@@ -11,9 +11,11 @@ categories:
 This is the list of reserved words you shouldn't use as field names, in addition to the names of Statamic's [Tags](/tags) and [contextual variables](/variables).
 
 - `content_type`
+- `count`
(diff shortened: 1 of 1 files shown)


duncanmcclean:
I have a feeling we might be able to make `hook` fields *just work*, without needing to add it to the list of reserved words.

I haven't tested it but we might be able to add `hook` to the array here: https://github.com/statamic/cms/blob/5.x/src/Data/AbstractAugmented.php#L95

duncanmcclean (approved):
Thanks for this pull request!

I've removed the addition of `hook` to the list since it shouldn't need to be a reserved word. I've opened a PR in Statamic to fix it: https://github.com/statamic/cms/pull/10319.

I've kept the alphabetical recording though 🙂

PR #1366, opened 2024-06-12
Update example of custom invalidator for static caching

godismyjudge95:
I found some of the things in the custom invalidator example to be confusing.  This is a PR to clarify and also provide a complete working example.


duncanmcclean (approved):
Thanks!

PR #1371, opened 2024-06-14
Edit docs for storing users in a database to place instructions in correct order

steveparks:
In https://statamic.dev/tips/storing-users-in-a-database :

I suggest moving the tip about composer require doctrine/dbal to a position in the docs where it is most useful. 

Otherwise the reader works down all the instructions, and then finds this when it's too late!


duncanmcclean:
Out of curiosity, which Laravel version were you running these steps on? 

Laravel 11 (and for the most part Laravel 10 too) removed their dependance on `doctrine/dbal`, so this hot tip might not be needed anymore.

steveparks:
Aha! I'm on Laravel 11.10.0 and hadn't observed any issues by the time I got to the end of this page, but then followed this instruction anyway, and thought it'd be more useful further up the page. But yes, maybe it could just be removed.

duncanmcclean (approved):

duncanmcclean:
I've just removed the suggestion completely - thanks!

PR #1372, opened 2024-06-16
Update contribution-guide.md

heidkaemper:
Updated shell output and screenshots to match Statamic 5 and Vite.


content/collections/docs/contribution-guide.md
```diff
@@ -190,7 +194,7 @@ git checkout -b fix/issue-9999
 When requiring the `cms` package, it's important to require the appropriate constraint. If you don't use the right one, Composer may decide to use the _real_ `cms` package, and you'll be left wondering why your code changes aren't appearing.
 :::
 
-If the branch is numeric then you need to require `BRANCH.x-dev` (e.g. a branch named `3.2` should use a constraint of `3.2.x-dev`).
+If the branch is numeric then you need to require `BRANCH.x-dev` (e.g. a branch named `5.7` should use a constraint of `5.7.x-dev`).
```
duncanmcclean: Since we're now using semantic versioning, this should be `5.x` instead:

```suggestion
If the branch is numeric then you need to require `BRANCH.x-dev` (e.g. a branch named `5.x` should use a constraint of `5.x-dev`).
```

duncanmcclean (changes requested):

duncanmcclean (approved):
I've made a couple of small tweaks, but looks good! Thanks.

PR #1378, opened 2024-06-20
Remove whereCollection and whereInCollection from docs

j3ll3yfi5h:
Methods are not available anymore and throw errors:

Call to undefined method Statamic\Stache\Query\EntryQueryBuilder::whereCollection()

Call to undefined method Statamic\Stache\Query\EntryQueryBuilder::whereInCollection()

### DIFF
--- a/content/collections/repositories/entry-repository.md
+++ b/content/collections/repositories/entry-repository.md
@@ -25,8 +25,6 @@ use Statamic\Facades\Entry;
 | `findByUri($uri, $site)` | Get Entry by `uri`, optionally in a site |
 | `findOrFail($id)` | Get Entry by `id`. Throws an `EntryNotFoundException` when the entry cannot be found. |
 | `query()` | Query Builder |
-| `whereCollection($handle)` | Get all Entries in a `Collection` |
-| `whereInCollection([$handles])` | Get all Entries in an array of `Collections` |
 | `make()` | Makes a new entry instance |
 
 ## Querying


duncanmcclean:
The [`whereCollection`](https://github.com/statamic/cms/blob/d299cb064d32802da9e2c83f73b2b419c884e202/src/Stache/Repositories/EntryRepository.php#L36-L43) and [`whereInCollection`](https://github.com/statamic/cms/blob/d299cb064d32802da9e2c83f73b2b419c884e202/src/Stache/Repositories/EntryRepository.php#L45-L52) methods *do* still exist on the `Entry` repository/facade. 

However, by the looks of those error messages, you're trying to use them on the query builder (where they don't exist)?

j3ll3yfi5h:
@duncanmcclean Ah interesting, you're right!

Can you maybe explain to me, what is the difference? Most examples on this page https://statamic.dev/repositories/entry-repository are about using the methods in query builders, right? Or am I completely confused?

duncanmcclean:
So, the methods here are the methods on the `Entry` facade (& the repository, which is what the facade hits when you call it):

CleanShot 2024-06-20 at 14 07 37

The `query` method will give you a "query builder" instance. The query builder is another object which lets you query & order entries. It's what lets you do filtering like `->where('foo', 'bar')`.

The query builder has a different set of methods to those on the repository. 

For example:

```php
Entry::query() // the query method makes a "query builder"
    ->where('collection', 'pages') // adds a "condition" to the query, filtering entries to those where the collection = pages
    ->where('foo', 'bar') // adds another "condition" to the query, filtering entries to those where foo = bar
    ->get() // actually executes the query to get the results. the get method returns a Laravel Collection instance
    ->map() // this map method exists on the Laravel Collection instance (which you can find out more about here: https://laravel.com/docs/master/collections)
```

Does that help?

j3ll3yfi5h:
@duncanmcclean Aha, I understand. Thank you very much for your explanations!

PR #1383, opened 2024-06-27
Update search.md

robdekort:
Those settings do not seem to belong in a settings array. They don't work when they are.


jackmcdade:
Good catch, thanks!

PR #1386, opened 2024-06-28
Controllers: cascade entry content

husseinalhammad:
This PR updates the Controllers page to include an explanation of how to cascade an entry's content into an Antlers view with the `cascadeContent` method.


duncanmcclean (approved):
Thanks!

PR #1393, opened 2024-07-02
Fix small typos in views.md

enderandpeter:
# Problem

There's a small typo on the [views](https://statamic.dev/views) page

# Solution

This PR fixes the typo

# Notes

Thanks for accepting these PRs 🙂

[Don't forget](https://www.grammarly.com/blog/its-vs-its/) that **its** is always a possessive adjective and **it's** is always a contraction for "it is" (or "it has")


jackmcdade:
Thank you!

PR #1396, opened 2024-07-04
Update static-caching.md

robdekort:
By making a small example difference we now explain that you it is possible to invalidate `all` taxonomies, globals or navigations instead of having to write them all out.


duncanmcclean (approved):
Thanks!

PR #1402, opened 2024-07-08
Update docker install guide to not need a custom docker-compose.yml

jamiel:
Removes the need to maintain a custom `docker-compose.yml` file in the docs.

The suggested `docker-compose.yml` file is currently out of date as it is missing the Vite ports and also uses php 8.2 rather than 8.3 which is what the latest Laravel build script would use.

In addition the current suggested build script for Mac OSX users sets all the defaults in the Laravel config to use MySQL so when switching to the version without MySQL there are additional configuration steps needed to get it working which aren't currently in the documentation.

So this change should make it easier for most users to get up and running if they just follow the standard new Laravel project instructions which defaults to sqlite.

### DIFF
--- a/content/collections/docs/docker.md
+++ b/content/collections/docs/docker.md
@@ -23,45 +23,13 @@ If you don't already have Docker installed, head to [docker.com/get-started](htt
 
 ## Installing Laravel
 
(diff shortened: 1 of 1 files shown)


jackmcdade:
Sounds good to me, thank you!

PR #1403, opened 2024-07-08
Add warning about using existing tag names for taxonomies

steveparks:
At the moment there is no disambiguation between tags for taxonomies and tags for system tags within views.

So a taxonomy called 'Section' can't be referenced as ```{{ section }} <li> ... </li> {{ /section }}``` in views because it conflicts with the existing Statamic system tag {{ section }}.

There is a way to disambiguate between variables and tags, but not between types of tags.

I've submitted an idea to provide this disambiguation: https://github.com/statamic/ideas/issues/1195

But in the meantime, I suggest adding this warning to the taxonomy page to avoid people getting in a pickle.

### DIFF
--- a/content/collections/docs/taxonomies.md
+++ b/content/collections/docs/taxonomies.md
@@ -21,6 +21,11 @@ Each entry in a taxonomy is often called a **term**.
 Watch how to set up your first Taxonomy
 :::
 
+> [!CAUTION]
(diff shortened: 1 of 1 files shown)


sjclark:
To be fair Statamic does advise not to use the names of `tags` as field names, but maybe that should be expanded to be a bit broader (beyond just field names). Generally using the same names in various places is probably going to break something eventually 😂  https://statamic.dev/tips/reserved-words

PR #1427, opened 2024-08-05
Update encrypt cookies docs

jacksleight:


content/collections/tags/cookie.md
```diff
@@ -83,4 +83,4 @@ You can remove cookies by passing the names of the variables into the `keys` par
 
 By default, in Laravel, cookies are encrypted so you will not be able to access the values of any data you set outside of PHP.
 
-To prevent encryption you need to add an exception to the `$except` array in your `app/Http/Middleware/EncryptCookies.php` file.
+To prevent encryption you need to [add an exception](https://laravel.com/docs/11.x/responses#cookies-and-encryption) to the EncryptCookies middleware.
```
duncanmcclean: Instead of linking out to the Laravel docs here, it might be better if we add an example on this page.

Also, since we still support Laravel 10, we'll need to keep the "add an exception to EncryptCookies.php" bit.

duncanmcclean (changes requested):

content/collections/tags/cookie.md
jacksleight: Ah yeah, good point. I will tweak.

duncanmcclean (approved):
Thanks!

PR #1428, opened 2024-08-05
Update ubuntu.md

AlexJFox:
Add troubleshooting to avoid 502 errors if correct version of php-fpm isn't specified in the default config


duncanmcclean (approved):
Thanks!

PR #1432, opened 2024-08-08
Update Widget documentation, only FIRST letter of widget class can be capitalized

paulpreibisch:
The docs need to be updated, since there is a serious bug where the class name must only have first letter capitalized.


marcorieser:
Wouldn't it make more sense to fix the bug instead of mentioning in the docs that there is a bug?

duncanmcclean:
I don't think it's a bug, but I think `examplewidgetname` was a bad example.

I've updated the example in ad93ee8.

PR #1456, opened 2024-09-19
Dynamic asset folders

jasonvarga:
Goes with https://github.com/statamic/cms/pull/10808


sjclark:
:smile: Does the docs search index need flushing or something - took me a while to find this as it's not indexed (for me)?

Screenshot 2024-09-27 at 9 21 25 am

duncanmcclean:
We're using Algolia Docsearch, so we can't trigger a reindex ourselves, it does it on a schedule. It should show up in the next few days.

You can view the docs here: https://statamic.dev/fieldtypes/assets#dynamic-folders

PR #1467, opened 2024-10-01
Use the latest version of Laravel in docs links

osbre:
I also manually verified that each link works. Some left as before due to the section no longer existing in the newer version.

Because Laravel 10 is [EOL](https://endoflife.date/laravel) in 4 months, I replaced its links to 11 as well.


duncanmcclean (approved):
Thanks, Ostap!

PR #1472, opened 2024-10-04
Add section about the live_preview tag in antlers.

DanielDarrenJones:


duncanmcclean:
Would you be able to add it to the Variables section as well/instead?

https://statamic.dev/reference/variables

DanielDarrenJones:
> Would you be able to add it to the Variables section as well/instead?
> 
> https://statamic.dev/reference/variables

Yeah I can add that as well, will sort later on today / tomorrow 👍

duncanmcclean (approved):
Thanks for this - just moved it to the "Variables" section.

PR #1476, opened 2024-10-07
Update search.md

geertjanknapen1:
Adds documentation for the `--all` option. Also corrects the fact that `search:update` does in fact not update all indexes but will show the user a prompt from which they can select which indexes to update. All is also an option in the prompt.


duncanmcclean (approved):
Thanks!

PR #1492, opened 2024-10-15
Specify Algolia client version

jacksleight:
The Algolia driver does not currently work with version 4 of the Algolia client package (results in a `Class "Algolia\AlgoliaSearch\SearchClient" not found` error), but if you follow the current docs that's what gets installed.

This PR updates the docs with the latest supported version.


duncanmcclean (approved):
Thanks!

PR #1503, opened 2024-10-22
Fix "Get a User Group" example

joshtaylordev:
The example for Get a User Group isn't correct, and returns the following error:
`Too few arguments to function Statamic\Auth\UserGroup::get()`

It should just be `User::find('admin');`


duncanmcclean (approved):
Thanks!

PR #1511, opened 2024-10-24
Updates to security documentation

palmiak:
Based on [this Twitter conversation](https://twitter.com/palmiak_fp/status/1848847714764628013) I added some changes to the security section. 

I used the term "around 5%" but we can go with "many" for a better marketing vibe ;) 

Disclaimer - I work at Patchstack, but I'm adding this report not because I work there but because Sucuri focuses more on malware (and malware shows up on your website after using the vulnerability) while we focus on vulnerabilities. So both our reports assert a bit different moment in live of a hacked website :)

### DIFF
--- a/content/collections/docs/from-wordpress-to-statamic.md
+++ b/content/collections/docs/from-wordpress-to-statamic.md
@@ -118,9 +118,9 @@ And if you need more customization, addons like [Flexible Forms](https://statami
 
 ## Security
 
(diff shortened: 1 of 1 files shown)


jackmcdade:
Very nice, thank you!

palmiak:
That fast - WOW :D 

Thanks

PR #1519, opened 2024-10-28
Document support for event listener auto-registration from typehinted __invoke methods

### LINKED ISSUE #1515: Document changes to add-on listener auto-registration
I've opened a PR against the CMS (https://github.com/statamic/cms/pull/11009) that changes the behaviour of add-on event listener auto-registration as documented here: https://statamic.dev/extending/addons

I'd suggest that the Tip at the top of that page is updated to add 'or __invoke' as per the changes in bold below if that PR is accepted:

"As long as your event listener lives in src/Listeners and the event is typehinted in the listener’s handle **or __invoke** method, they will be automatically registered by Statamic, without you needing to register them manually."

leewillis77:
Fixes #1515 

Open to suggestions if the docs need to note a version number (5.34.0?) when __invoke() support was added?

### DIFF
--- a/content/collections/extending-docs/addons.md
+++ b/content/collections/extending-docs/addons.md
@@ -443,7 +443,7 @@ return view('custom::foo');
 ::: tip
 Statamic v5.33.0 introduced the concept of "autoloading" for event listeners.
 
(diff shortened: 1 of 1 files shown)


duncanmcclean:
Thanks!

It doesn't look like I can push to your branch so I'll tweak the version note after merging (it's kind of awkward). 👍

duncanmcclean:
Whoopsie, I merged this too early 🙈. Replacing with #1520.

PR #1520, opened 2024-10-28
Document support for event listener auto-registration from typehinted `__invoke` methods

duncanmcclean:
Whoopsie, I merged #1519 before its associated Core PR has been released. 🤦‍♂️

I've reverted that PR on `master` and opened this PR which'll replace it. It also changes the version mentioned in the docs to v5.35.0, which includes some fixes for autoloading event listeners.


duncanmcclean:
Merging now the associated Core PR has actually been tagged.

PR #1521, opened 2024-10-28
Document new form tags. 

### LINKED ISSUE #1516: Documentation for new Form tag hooks
The PR at https://github.com/statamic/cms/pull/11010 introduces three new hooks into Form tag processing which should be documented on the available hooks page (https://statamic.dev/extending/hooks#available-hooks)

leewillis77:
This PR adds documentation for the three new form tag hooks introduced in https://github.com/statamic/cms/pull/11010 and fixes #1516.

There doesn't seem to be any obvious "order" to the existing listing, so I've just added these after the Collection tag, however if you'd like me to re-order the existing items alphabetically or similar, happy to update this PR to do that as well, e.g.

- All tags
- Augmentation
- Bard items
- Collection
- Entry index query
- Form
- Static cache warming


jasonvarga:
Thank you

PR #1522, opened 2024-10-28
Update from-wordpress-to-statamic.md

bealers:
Great guide; I just felt some mention of Starter Kits would be helpful. 

I put it quite high on the page on purpose as I feel it would be something front of mind to someone considering moving to Statamic from WP. Cheers :)


jackmcdade:
Great idea, thank you!

PR #1531, opened 2024-11-05
Uninstalling addons

jasonvarga:
Goes with statamic/cms#11057


duncanmcclean:
Should we add the `pre-package-uninstall` script into the install steps for [an existing Laravel app](https://statamic.dev/installing/laravel#install-statamic)?

CleanShot 2024-11-06 at 12 28 16

PR #1540, opened 2024-11-08
Undocument cache tagging

duncanmcclean:
This pull request removes reference to Laravel's cache tagging functionality in our `{{ cache }}` tag, since Laravel now discourages the use of it.

It's no longer included in the Laravel docs, see https://github.com/laravel/docs/commit/63dde39470588ab1ad39cea9a592f0838ca5ad47.

There's no comment from the Laravel team on that commit but Taylor discussed it briefly during a Laracon AU Q&A and discourages people from using it.


jasonvarga:
I don't think it even worked exactly how we documented it either.

PR #1548, opened 2024-11-13
Fix Highlighting #1545

### LINKED ISSUE #1545: Highlighting in dark mode is unreadable
grafik

Using Firefox 132

andjsch:
Fix #1545

This PR fixes the highlighting not only in dark mode, but across all browsers. Currently it has only been working in Firefox as of my testing.

### DIFF
--- a/resources/css/content.css
+++ b/resources/css/content.css
@@ -1,12 +1,8 @@
 /* Main Content */
 /* -------------------------------------------------- */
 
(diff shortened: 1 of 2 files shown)


edalzell (commented):
Really hard to see what you changed here @andjsch because of the formatting changes. Can you please revert those so only your change is in here?

PR #1552, opened 2024-11-19
Field actions

jasonvarga:
Goes with statamic/cms#10352 and statamic/cms#11129


content/collections/extending-docs/field-actions.md
```diff
@@ -0,0 +1,209 @@
...
+                instructions: 'Enter your name',
+                validate: ['required', 'min:2']
+            },
+            color: {
+                type: color,
+                instructions: 'Select the color',
+            }
+        }
+    },
+    run: ({ confirmed }) => {
```
jacksleight: Should be `confirmation`

PR #1553, opened 2024-11-21
PHP 8.4

duncanmcclean:
Can be merged after this PR has been tagged: https://github.com/statamic/cms/pull/11114


jasonvarga:
Let's avoid suggesting 8.4 for the moment since there are deprecation warnings we can't address until v6. It _works_ but is a little meh.

PR #1558, opened 2024-11-24
Document new `default` and `skip_option` behavior

jesseleite:
For when https://github.com/statamic/cms/pull/11045 gets tagged.


duncanmcclean:
I'm assuming this is good to go - merging! 🚀

PR #1559, opened 2024-11-26
Prevent default when pressing search hotkeys

andjsch:
Prevent the default action when pressing ``cmd + k``. Within Firefox, this focuses the URL bar which is an unwanted side effect that should not happen. With this change, it focuses the search bar of the docs as intended.


duncanmcclean (approved):
Thanks!

I wish GitHub did this for their command pallette 😅

PR #1560, opened 2024-11-27
Fix Replacer interface class in docs

maxsolomon14:
I think the class referenced in the docs for Replacers shouldn't be the `NoCache` replacer it should be `Statamic\StaticCaching\Replacer` instead.


duncanmcclean (approved):
Thanks!

PR #1564, opened 2024-12-01
add width variable for fields

efc:
Using Statamic 5.40 and discovered `width` was available, which is important if you want to create a form that accurately honors the field widths assigned in the form blueprint.


duncanmcclean (approved):
Thanks!

PR #1565, opened 2024-12-03
Fix Mangled Headings

JohnathonKoster:
Bumps the min. version of the underlying search provider. The previous version was skipping headings if they did not contain leading content, but contained content inside nested headings:

```

## A heading

### A sub heading

The "A heading" would be skipped.

```

This was causing some search phrases like `templating` to act strangely when building up the parent headings


jackmcdade:
Thank you!

PR #1572, opened 2024-12-12
Update forge docs to recommend disabling toggles

Technobabble17:
A more standard install might recommend not running migrations or composer installing during the install repo stage.

I added a best practice tip, and updated the image with a current and compressed snapshot.

I didn't include anymore, but it might be great to discuss how to setup the environment variables afterwards as well as generating the deploy key!


jasonvarga:
Why would you need to disable those?

Technobabble17:
Using statamic v5 and cli 3.4.1, when installing a repo in Forge, the install will error out, even when setting the environment tab ahead of time. 

```
...
Generating optimized autoload files
> Illuminate\Foundation\ComposerScripts::postAutoloadDump
> @php artisan package:discover --ansi

In Connection.php line 825:
                                                                               
  Database file at path [/home/forge/forge-test.com/database/database.sqlite]  
   does not exist. Ensure this is an absolute path to the database. (Connecti  
  on: sqlite, SQL: select exists (select 1 from sqlite_master where name = 's  
  ites' and type = 'table') as "exists")                                       
                                                                               
In SQLiteConnector.php line 37:
                                                                               
  Database file at path [/home/forge/forge-test.com/database/database.sqlite]  
   does not exist. Ensure this is an absolute path to the database.            
                                                                               
Script @php artisan package:discover --ansi handling the post-autoload-dump event returned with error code 1
```

duncanmcclean:
You shouldn't need to disable the composer checkbox. 

Although, fair enough on the database toggle. You only really need it if you're using the Eloquent Driver.

duncanmcclean:
I'm still not sure you should need to uncheck the composer checkbox, but I've covered the database migrations issue in the deployment script section.

PR #1574, opened 2024-12-15
SEO Pro Docs

JohnathonKoster:
Goes with https://github.com/statamic/seo-pro/pull/348


jackmcdade:
This would be pretty stale by now. Closing. Can always revisit.

PR #1586, opened 2025-01-08
Use Laravel's resource_path instead of __DIR__

Humni:
Minor improvement

### DIFF
--- a/content/collections/extending-docs/fieldtypes.md
+++ b/content/collections/extending-docs/fieldtypes.md
@@ -144,7 +144,7 @@ class CustomFieldtype extends Fieldtype
     // or
     function icon()
     {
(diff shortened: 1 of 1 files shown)


marcorieser:
@Humni Are you sure the wrapping `file_get_contents()` is not needed?

Humni:
@marcorieser Yes - I have it running fine (unless I'm missing something?)

It's the recommended way to get the fully qualified resource path in laravel - https://laravel.com/docs/11.x/helpers#method-resource-path

Humni:
@marcorieser yeah no you are right there - I've updated my local example (thought it was working but it wasn't sorry!). 

Will update the PR.  The `__DIR__` from the old version didn't actually work for me out of the box, hence the change in the first place.

joshuablum:
Thanks!

PR #1589, opened 2025-01-17
Fix: Assets repositories docs links

MrMohamedAbdallah:
Links for `AssetsContainer` and `Assets` in these two pages were wrong.


duncanmcclean (approved):
Thanks!

PR #1591, opened 2025-01-20
Update variables page

miicah:
Added the partial to submit changes to GitHub and better clarified how to use front-end matter.


duncanmcclean (approved):
Thanks!

PR #1595, opened 2025-01-22
Launch Local Site With Please

mikeritter:
I changed the CLI command to use please instead of artisan to launch local server because artisan is not included.


duncanmcclean:
> I changed the CLI command to use please instead of artisan to launch local server because artisan is not included.

Artisan should be included in Statamic apps - it is on our [`statamic/statamic`](https://github.com/statamic/statamic/blob/5.x/artisan) starter app.

PR #1601, opened 2025-01-28
Update children.md

skoontastic:
Add documentation for checking the current page's children variable.


duncanmcclean:
You can also use a $ dollar sign to ensure you're working with the children variable, and not the children tag. See [the Antlers docs](https://statamic.dev/antlers#disambiguating-variables):

CleanShot 2025-04-07 at 10 41 52

PR #1605, opened 2025-02-06
Using variable keys with GET

Vyygir:
This took me a while to figure out how to use appropriately in a project so I thought I'd submit something for the docs as a notation of it

If it's not perfect, all good, at least it's flagged now if someone wants to write something better than I have ✨


duncanmcclean:
I don't know that we need to document it specifically for the `get` variable, it's just [how Antlers works](https://v6.statamic.dev/antlers#dynamic-access).

PR #1607, opened 2025-02-07
Update zero-downtime-deployments.md

gregwolanski:
Fixes outdated default branch name. I know it's a small thing, but I wasted a few minutes after copying and pasting the snippet myself. I think it's worth removing that friction for other people trying Statamic in the future.

[Context](https://github.com/github/renaming)


duncanmcclean (approved):
Thanks!

PR #1611, opened 2025-02-14
FIX: missing text color weight

krsriq:
Followed the quick start guide and was confused for a sec why my nav items were not showing: there is a small bug in the example code (missing tailwind text color weight), this fixes it.


duncanmcclean:
Thanks!

PR #1619, opened 2025-02-17
Fix typos

krsriq:
The next batch of typo fixes.


jackmcdade:
Wow, thank you so much. That was so many in one shot that I'm quite embarrassed. 😅

PR #1620, opened 2025-02-18
Added note about how to enable localizable fields in multi-site

mytchallb:
Added a note about how to enable field localization


joshuablum:
Thanks!

PR #1626, opened 2025-02-19
Update trans tag

andjsch:
Updating the translation files location. I think this one could be merged together with L12, even though it's true for L11 already.


duncanmcclean (approved):
Thanks!

PR #1633, opened 2025-02-27
fix: broken link in footer

JorisOrangeStudio:
See [this link](https://statamic.dev/#:~:text=Partners-,Become%20a%20Partner,-Search) in the footer
Should be https://statamic.com/partners/join instead of https://statamic.com/partners/how-to-join


duncanmcclean (approved):
Thanks!

PR #1635, opened 2025-02-28
Fixed sample code in importing-content.md

fgd007:
array in sample code wasn't correct. fixed the key => value mapping


duncanmcclean (approved):
Thanks!

PR #1638, opened 2025-03-04
Fixed missing comma in example on where.md

ELowry:
Passing operators to the where modifier must be comma-separated from the value.


duncanmcclean (approved):
Thanks!

PR #1649, opened 2025-03-27
Minor fix blade tag {{ current_url }}

stoffelio:
Fixing blade tag (uri -> url)

Looks like this was copied from the {{ current_url }} tag and didn't get changed.


duncanmcclean (approved):
Thanks!

PR #1650, opened 2025-03-31
Update reserved-words.md

j3ll3yfi5h:
add `date` handle because it breaks the validation message
Bildschirmfoto 2025-03-31 um 17 16 15


duncanmcclean:
You _should_ be able to use `date` as a field handle, even without it being a dated collection. If you're running into an issue, please [open an issue](https://github.com/statamic/cms/issues).

j3ll3yfi5h:
@duncanmcclean done ✅

PR #1658, opened 2025-04-09
Fixes docs referencing migrating users from flat-file to database

Code-Derek:
The database model for user roles and groups was updated in [this commit](https://github.com/statamic/cms/commit/96aa7eaa9db0545c2f60cc3042bfeffe4e397472).

The current documentation instructions for adjusting the migration to use UUIDs instead of incrementing IDs is outdated compared to the migration generated by `php please auth:migration`


duncanmcclean (approved):
Well spotted, thanks!

PR #1660, opened 2025-04-10
Update Markdown example to reflect the new extensions added to core

tao:
A recently merge pull request [#8417](https://github.com/statamic/cms/pull/8417) added some additional Markdown extensions to the core.

The documentation has been updated to include the new additions, and update the example to include the HintExtension instead of the FootnoteExtension (included by default now) as that might cause some confusion, seen in this bug [#8495](https://github.com/statamic/cms/issues/8495).


duncanmcclean (approved):
Good catch, thanks!

PR #1672, opened 2025-04-29
Update regex_replace.md

efc:
It seems we should really have an example for `regex_replace` that uses a regular expression substitution. This is just one I have on hand, but any regular expression substitution should do.

### DIFF
--- a/content/collections/modifiers/regex_replace.md
+++ b/content/collections/modifiers/regex_replace.md
@@ -9,21 +9,21 @@ title: 'Regex Replace'
 Run a find and replace regex on a string of content.
 
 ```yaml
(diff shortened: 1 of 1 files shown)


jackmcdade:
Closing in favor of 79d0a67fa48954b23cdf588d2bb1e6ccdfd2b184. Thanks for the suggestion!

efc:
Almost, @jackmcdade... unfortunately the result shown does not match the regex in the example.

marcorieser:
> Almost, @jackmcdade... unfortunately the result shown does not match the regex in the example.

What's not working? seems correct to me: https://run.antlers.dev/5j8Ytq0Yt9Vh0XWO

PR #1679, opened 2025-05-12
Clarification concerning computed values's limitations

StevenRiot:


duncanmcclean (approved):
Thanks! I've tweaked the wording slightly.

You _should_ be able to use computed fields from fieldsets - if you're having issues with that, please open an issue.

StevenRiot:
I actually wasn't sure about that part. I tested it a few months ago and didn't remember exactly but I figured you'd correct me.

StevenRiot:
@duncanmcclean I actually found a workaround. The page needs to be saved or refreshed everytime you add a new item to your replicator for the value to be visble, but it works as intended.

In case you wanna add this to the doc, here's the process

The setup is the same as a top level field (field's visibility needs to be computed), then in your AppServiceProvider.php file : 

`Collection::computed('articles', 'replicator_handle', function ($entry) {
    $sets = $entry->get('replicator_handle', []);
    return array_map(function ($set) {
        $set['field_handle'] = 'foo';
        return $set;
    }, $sets);
});`

PR #1684, opened 2025-06-04
Updates conflict fields, Adds Assets, Navs and Trees to reserved list

adampatterson:
Adding `content` as a file driver disk to assets causes conflicts. 
Adding `structure` as an entry field name causes conflicts with the field, as well as accessing that data in a Nav Tree

### DIFF
--- a/content/collections/tips/reserved-words.md
+++ b/content/collections/tips/reserved-words.md
@@ -24,6 +24,7 @@ This is the list of reserved words you shouldn't use as field names, in addition
 - `save`
 - `site`
 - `status`
+- `structure`
 - `path`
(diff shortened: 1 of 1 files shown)


duncanmcclean:
> Adding content as a file driver disk to assets causes conflicts.

I know it's been a while since you opened this. Do you remember what kind of conflicts you ran into?

adampatterson:
@duncanmcclean I'd have to check my notes.

I think creating a structure entry sort of merged in different parts of the nav tree with the core structure.

I don't recall what what naming an asset container content did. I seem to think it was empty or would not display images after uploading them.

PR #1688, opened 2025-06-09
Static-caching: .htaccess snippet: Removed </Ifmodule> as there was no start-mention of this..

jeroenpeters1986:
Removed </Ifmodule> as there was no start-mention of this..
I think it would be logical to OR
 * add `<IfModule mod_rewrite.c>` to the top of the snippet
 * remove the `</IfModule>` because there was no start-IfModule

To have this </IfModule> might seem logical to most of us if you have .htaccess experience. If there was no start-tag, you will get an error on apache. Also, most hosting companies are providing Apache2 with mod-rewrite as most CMS'es required it these days.

Hope my comment makes some sense :-)


duncanmcclean (approved):
Thanks!

PR #1689, opened 2025-06-15
Collection tag (blade): from <collection:xxx> to <statamic:collection:xxx>

AtmoFX:
The 3 Blade examples provided for the collection tag page under [Aliasing](https://statamic.dev/tags/collection#alias), [Scoping](https://statamic.dev/tags/collection#scope) and [Grouping](https://statamic.dev/tags/collection#grouping) are all using `<collection:blog>` tag instead of `<statamic:collection:blog>` like the other examples.

This PR addresses this mistake.


duncanmcclean (commented):
Well spotted, thanks!

PR #1692, opened 2025-06-20
Update conditional-fields.md to include taxonomy example

### LINKED ISSUE #1450: Syntax for Conditional on Taxonomy Term not documented
Full discussion here: https://github.com/statamic/cms/discussions/6488

Most relevant excerpt:
I think this is the answer, though I'm not totally sure if these are the right identifiers. The secret sauce here missing from the docs is the `<some sort of handle>::`

```
if:
  <taxonomny_terms_field_handle>: 'contains <taxonomy_handle???>::<taxonmy_term_slug>'
```

So in my current example it looks like this:
```
unless:
  news_types: 'contains news_types::media-mentions'
```

efc:
Seems conditional fields based on taxonomy choices is a pretty common use case. Found this advice at https://github.com/statamic/cms/discussions/7546.

### DIFF
--- a/content/collections/docs/conditional-fields.md
+++ b/content/collections/docs/conditional-fields.md
@@ -137,6 +137,31 @@ If you are dealing with a string value, `contains` and `contains_any` will perfo
       favorite_food: 'contains_any pizza, lasagna'
 ```
 
+### Contains in a Taxonomy
+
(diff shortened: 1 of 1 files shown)


miicah:
Closes #1450

PR #1711, opened 2025-07-29
Fix the link to Antlers Toolbox

CasEbb:
The Antlers Toolbox link doesn't point to a landing page anymore, link to VS Code's Marketplace instead.


duncanmcclean:
Antlers.dev should still work. I'll ask John Koster take a look.

PR #1732, opened 2025-09-10
Update addons.md fixing the 'seller account' link

Braunson:
Adjusting seller account link in the Publishing to Marketplace section


duncanmcclean:
Thanks!

PR #1767, opened 2025-11-11
The dump tag does not stop execution

ryanmitchell:
Update the overview to remove this incorrect statement.


duncanmcclean:
Well spotted, thanks!

PR #1777, opened 2025-11-25
Add warning that caching needs to be disabled

lostgeek:
It is important to disable caching when using the GraphQL API with authorisation. Otherwise the authorised response will be served even to unauthorised clients.

See discussion on [discord](https://discord.com/channels/489818810157891584/489818810157891586/1441339700991365172).


jackmcdade:
Closing in favor of 3ad91f8d038761d931b30544172bc39d5157e2d9. Thanks for the suggestion!

PR #1810, opened 2026-02-02
Time field and modifier

jasonvarga:
This adds some extra information about the time fieldtype, and documents the new `format_time` modifier.

See statamic/cms#13799


duncanmcclean:
Do we need to mention this in the upgrade guide?

CleanShot 2026-02-03 at 09 45 26

jasonvarga:
> Do we need to mention this in the upgrade guide?

Isn't that exactly where that screenshot is from?

duncanmcclean:
Sorry, I meant mentioning the new modifier in the upgrade guide. The screenshot made that message confusing 😅

If someone updates to v6, and that option is enabled, wouldn't the `format` modifier suddenly output the wrong time? They'd need to adopt the `format_time` modifier instead?

jasonvarga:
No, the formatTime modifier parses the date value itself - it doesn't use the carbon() method, which is what uses that setting.

PR #1813, opened 2026-02-03
Merge extending Bard docs into Bard fieldtype docs

### LINKED ISSUE #1812: Extending bard doc has disappeard
https://v5.statamic.dev/extending/bard no longer exists in v6 doc. It redirects to https://statamic.dev/fieldtypes/bard, which has it's own section : https://statamic.dev/fieldtypes/bard#extending-bard with a link to ... itself !

joshuablum:
With our docs restructure both the Bard Fieldtype and Extending Bard docs share the same slug of `bard`. Pre restructure it was `fieldtypes/bard` and `extending/bard` but now it both lives in `fieldtypes/{page}`. So even though the Extending Bard page technically is still there you always see the Bard Fieldtype docs. Changed slug to `extending-bard` so it actually works again.

Closes https://github.com/statamic/docs/issues/1812.


duncanmcclean:
We should probably combine the pages instead.

PR #1827, opened 2026-02-13
Field Action fixes

duncanmcclean:
Depends on https://github.com/statamic/cms/pull/13921


HammenWS:
@duncanmcclean Please merge so the Statamic docs are correct

I wonder if there is more documentation that need an update.

PR #1865, opened 2026-03-20
Update examples for arrays and plucking in antlers.md

bastihilger:
I ran into this when trying out default content for variables in front matter (which is explained correctly here https://statamic.dev/tags/partial#passing-data. Before my change, it looked like the yaml part was actually a front matter part of the antler example - which would not work, since you would have to prefix those variables with `view:`.


duncanmcclean (approved):
Thanks!

PR #1910, opened 2026-05-09
tiny fix: Use $HOME variable in setting $PATH also on MacOS

damnkunze:
Easier to copy-paste and should always be correct, since this is default location for composer install.


duncanmcclean:
Thanks!

PR #1911, opened 2026-05-09
Correct is_after and is_before examples

efc:
OK, I went spelunking into the `isAfter` modifier code and discovered that the documentation is wrong. I think this is probably due to the adoption of Carbon dates (was Carbon adopted in v6?). The docs show `{{ if start_date | is_after(2014) }}`, which most readers would see as the year "2014", but the code treats this as "2014 seconds after the Unix epoch", which is quite different. In fact the code does this for any numeric value, whether provided as a number or a string. To get the expected result you would have to provide `is_after("2014-01-01")` or something of that sort. It turns out that `is_after("today")` does work, as does any valid Carbon relative date string, such as `-1 day` or `+2 months`.

This pull request aligns the docs with actual behavior. It does nothing to fix the behavior of numeric values, which someone might want to do in order to be backward compatible if v5 behaved as advertised.

### DIFF
--- a/content/collections/modifiers/is_after.md
+++ b/content/collections/modifiers/is_after.md
@@ -6,7 +6,7 @@ modifier_types:
   - conditions
 title: 'Is After'
 ---
-Returns `true` if a date variable is after another date. That second date can be the name of another variable or a literal date string.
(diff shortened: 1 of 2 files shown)


jackmcdade:
Thank you!

PR #1918, opened 2026-05-18
Vite tooling hint for addons

efc:
I have twice missed the very valuable information in the _Vite Tooling_ documentation because I was so focused on the _Building an Addon_ document. This pull request provides a bridge that might help future me (and others like me) who are so overwhelmed with the task of building an addon with a fieldtype that they miss important information in _Vite Tooling_.

This also provides the syntax for building fieldtypes into addons, which does not seem to be present in the docs otherwise.

### DIFF
--- a/content/collections/pages/building-an-addon.md
+++ b/content/collections/pages/building-an-addon.md
@@ -203,6 +203,19 @@ public function bootAddon()
 }
 ```
 
+## Including Fieldtypes with your Addon
+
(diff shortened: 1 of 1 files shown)


jasonvarga:
The "CSS and JavaScript" section on the "Building an Addon" page already links to the Vite tooling page.

efc:
Fair enough, Jason. If you all think that does the trick, I could understand that. I just know that I've missed that reference twice myself even when hunting for the solution to the "manifest not found" problem after installing a fieldtype into an addon. I just felt some focussed description of how fieldtypes are built in and the extra publish step would be helpful for others like me.

Issue #1946, opened 2026-07-02
Document OAuth Improvements


duncanmcclean:
https://github.com/statamic/cms/pull/14899

joshuablum:
I _think_ this got documented by https://github.com/statamic/docs/commit/b369b8977dd4d027f8a84c031da60b5644c9d4cc

PR #1947, opened 2026-07-09
Add trusted providers section to OAuth documentation

martinoak:
In release 6.24.2 and 5.74.1, a new trusted_providers config has been implemented. Since we use custom OAuth2 provider, people couldn't login into the Control Panel without adding the `trusted_providers` array with our provider in it.

Feel free to rewrite the docs, I just wanted to pointed out that this section is missing in the docs.

Related: #1946

### DIFF
--- a/content/collections/pages/oauth.md
+++ b/content/collections/pages/oauth.md
@@ -175,6 +175,12 @@ To create your own OAuth provider, you should make your own SocialiteProvider-re
 
 Follow the [third party installation steps](#third-party-providers), but skip the Composer bits. You can just keep the classes somewhere in your project.
 
+If you are using a custom provider and trust your own provider that the e-mails coming in are verified, you need to add `trusted_providers` with array of your own provider's handle:
+
+```php
+`trusted_providers => ['my-custom-provider']
+```
+
 ## Customizing user data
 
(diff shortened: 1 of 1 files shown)


jasonvarga:
@martinoak Logging in should've required no changes.

The only thing that would've changed would have been trying to logging into an **existing Statamic account** with an OAuth provider that was not already linked to the Statamic user.

Are you using v5 or v6?

martinoak:
@jasonvarga still on v5, we use our own company provider, but since the user wasnt even in the file that stores oauth id with our provider's identifier. When I added trusted_providers array key, everything worked even on clean installations.

jasonvarga:
But when the user tried to log in with your company OAuth for the first time, was there already a Statamic account with that same email?

martinoak:
Yes, otherwise "Unauthorized" message would appear.

jasonvarga:
Cool cool, thanks. Addressed in c3b2919e8ba6d48e2792c8b2276cc4e42fe78af6 for the v5 subdomain.
Equivalent for v6 coming shortly.

martinoak:
Just a small addon: Google is already in trusted_providers in vendor package, so repeating it here has no effect. I would add here the providers not listed in the core package for clarity.

jasonvarga:
Its defining all the providers, not just ones you want to add.

PR #1950, opened 2026-07-14
Break up the Users super page into smaller pages

jackmcdade:
Roles, groups, permissions, custom permissions, etc each get their own URL.


jackmcdade:
@duncanmcclean Wouldn't mind a sanity to check to make sure i didn't miss anything important splitting all these up.

duncanmcclean (approved):

PR #1953, opened 2026-07-15
Fix `inline: break` example

marcorieser:
The upgrade guide says `inline: break` should be changed. However the example uses `inline: inline`.


duncanmcclean:
Thanks!

PR #1959, opened 2026-08-05
Update bot detection logic in deployment scripts

espenlg:
To make sure that it will be detected in cases where Co-Author is a part of the commit message.


espenlg:
And then I noticed it was already mentioned that [BOT] should prepend the message. I guess both can work, it's just about where the change is done rather than if it will work.

PR #1974, opened 2026-08-25
Add Blade syntax for locales rendering

adevade:
Fixes bad rendering of Blade example on locales tag page.

https://statamic.dev/tags/locales#iterating-over-locales

image


duncanmcclean:
Thanks!

PR #1975, opened 2026-08-26
Modify middleware to include Content Security Policy

StephenThink:
Updated middleware configuration to add Content Security Policy to the web middleware group.

This is my first of doing something like this so I hope i have done it right, when I tried to do this I could not find the Kernel file and this is the reason for my update.

I am not sure if I understood the [tl! focus] correctly.


duncanmcclean:
Thanks!

I've just tweaked the indentation and merged this. The code annotations (`focus` and `++`) come from [Torchlight](https://torchlight.dev).

PR #1980, opened 2026-09-01
Fix env preprocessor

JohnathonKoster:
Adds the required `supports` method to the `StripEnvTrailingNewlines` preprocessor.


jackmcdade:
Thanks!

PR #1989, opened 2026-09-11
Remove JavaScript position section from nocache.md

bastihilger:
Removed section about JavaScript position for nocache tag because it is no longer needed: It was removed in v6 (https://statamic.dev/getting-started/upgrade-guide/5-to-6#full-measure-static-caching)


duncanmcclean:
Thanks!

